PR #36, opened 2023-04-14
minor edits to clarify examples and to reduce uses of terms like tidy or repository

harryhoch:
Minor changes

1. modified the first example on the overview page to indicate time horizon in weeks.

2. changed model-ouptut and target-data to minimize the use of the terms "tidy" and "repository" so as to not sound too platform/language dependent.


annakrystalli (approved):
LGTM

PR #43, opened 2023-04-24
adding changes to document modeling tasks more clearly on its own page

nickreich:
maybe not entirely finished, but a first pass at making a new page that defines more specifically the modeling tasks.


annakrystalli (approved):
Looks great!

PR #44, opened 2023-04-24
Update Hub config page

LucieContamin:
- rename hub metadata to hub config
- add more/update information on Hub config files


annakrystalli (approved):
All good

PR #47, opened 2023-06-07
Misc updates

nickreich:
changes throughout the docs to address various small issues.


docs/source/format/tasks.md
```diff
@@ -30,7 +30,9 @@ The figure shows that Hub metadata and target data are specified by the hub itse
...
+We note that some task ID variables are special in that they conceptually define a modeling "target" (these are referred to in the [tasks metadata](tasks-metadata) as a `target_key`). In our [Running Example 1](running-examples), the task ID variables are `target`, `location`, `origin_date`, and `horizon`. In this example, `target` is the target key. In other examples, (such as [Running Example 3](running-examples)) more than one variable can serve as target keys together.
+
+Some task ID variables serve specific purposes. For example, every hub must have a single task ID variable that uniquely defines a submission round. It has become a convention to use a task ID like `origin_date` or `forecast_date` for this purpose, although in practice hubs could use other task ID variables for this  purpose. In the  [Running Example 1](running-examples), this column is `origin_date`. Additionally, for hubs with step-ahead short term predictions, to plot forecasts using standard functionality in the `hubUtils` package, hubs must have either `origin_date` and `horizon` columns or a `target_date` column.
```
annakrystalli, line 35:
```suggestion
Some task ID variables serve specific purposes. For example, every hub must have a single task ID variable that uniquely defines a submission round. It has become a convention to use a task ID like `origin_date` or `forecast_date` for this purpose, although in practice hubs could use other task ID variables for this  purpose. In  [Running Example 1](running-examples), this column is `origin_date`. Additionally, to plot forecasts using standard functionality in the `hubUtils` package in hubs with step-ahead short term predictions, hubs must have either `origin_date` and `horizon` columns or a `target_date` column.
```

annakrystalli: In addition to the minor text edit suggestion above I had two questions:

1. Should we be referring to columns or task IDs in:
    > hubs must have either `origin_date` and `horizon` columns or a `target_date` column.
2. Should be it be `origin_date` and `horizon` columns or a `target_date` column or **`origin_date` and `horizon` columns or origin_date` and `target_date` columns**. I _thought_ discussions concluded that there should always be an `origin_date` task ID/column but maybe I misunderstood?

annakrystalli (commented):
LGTM overall. Some minor suggestions and requests for clarifications added

docs/source/format/tasks.md
nickreich: 1. I agree. Pushed a change.
2. Regarding this, I think the conclusion was that there needs to be a round identifier column but that it need not be `origin_date` necessarily. So therefore, it's better as is I think.

nickreich:
Per verbal approval from @annakrystalli , merging in.

Issue #82, opened 2023-12-21
 No mention of using explicit end & start date when configuring "submission_due"  on page /quickstart-hub-admin/tasks-config.html


annakrystalli:
In the [Setting up `"submission_due"`](https://hubdocs.readthedocs.io/en/latest/quickstart-hub-admin/tasks-config.html#setting-up-submissions-due) section, there is no mention of setting up rounds with explicit end & start date submission windows instead of submission windows relative to a reference date.

See schema for details:

https://github.com/Infectious-Disease-Modeling-Hubs/schemas/blob/de580d56b8fc5c24dd36a32994182e37b8b0ac95/v2.0.0/tasks-schema.json#L1323-L1380

micokoch:
Explained this more clearly in [PR#82](https://github.com/Infectious-Disease-Modeling-Hubs/hubDocs/pull/84). If that seems okay to you (and once the PR is merged), we can close this issue.

PR #84, opened 2024-01-11
Code color in dark/light, email & Issue #82

micokoch:


docs/source/overview/contact.md
```diff
@@ -1,7 +1,7 @@
 # Contact  
 
 - If you would like to learn more about our work and join our open, monthly community meetings, please [join our email listserv](https://groups.io/g/hubverse) where we send occasional updates (1-2 times per month) on our activities and the agenda for our monthly meetings.  
-- To reach out directly, please send a message to <a href="mailto:hubverse+owner@groups.io">the group owner</a>.  
+- To reach out directly, please send a message to <a href="mailto:mzorn@umass.edu">the group owner</a>.
```
nickreich, line 4: I think we should leave this as the more generic hubverse+owner email. otherwise we'll have to rememeber to update this when Martha leaves, etc...

docs/_static/css/custom.css
```diff
@@ -52,9 +52,17 @@ figcaption {
   text-indent: 2em;
 }
 
-code.literal {
-    color: #660066 !important;
-    background-color: inherit !important;
+html[data-theme="light"] {
```
nickreich, line 55: These colors seem fine to me. I will note that the logo colors are here, in case there is a way to thematically choose colors here or elsewhere on the site. 

Blue - HEX #3C88BE; R(60), G(136), B(190); C-76, M-37, Y-8, K-0
Light Blue - HEX #7AC6F4; R(122), G(198), B(244); C-53, M-7, Y-0, K-0

docs/source/overview/contact.md
nickreich: For example, as least I receive emails sent to that above address, possibly others too.

nickreich (changes requested):
I suggest not making the email change. The colors seem ok.

docs/source/overview/contact.md
micokoch: Hi Nick. I wasn't sure if anyone checked the "group owner" email. I changed all references to Martha's email to go to the "group owner" email.

docs/_static/css/custom.css
micokoch: I can change all colors (I think), but since there are only two colors in this palette, it makes sense to limit them to very specific uses. Since I only set default colors for 3 things (using custom css), I decided to use them for those three uses:
1. I used the blue for image borders.
2. I used the blue for `code` in the light mode.
3. I used the light blue for `code` in the dark mode.
The downside to this is that the blues resemble hyperlinked text, so I can revert if you prefer.

nickreich (approved):
All looks good to me!

PR #90, opened 2024-02-29
Update docs for cloud-enabled schema (v2.0.1)

### LINKED ISSUE #89: Add documentation for optional configuration of a cloud backend
As part of the [Hubverse Cloud project](https://github.com/orgs/Infectious-Disease-Modeling-Hubs/projects/5/views/1), update Hubverse documentation to include the following:

- [ ]  Background/overview of what a cloud-enabled hub does and how it works
- [ ] How to set up admin config for cloud-enabled hubs

bsweger:
Resolves #89 

Bump the schema version and add some information about cloud-enabled hubs.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -13,7 +13,7 @@ Jinja2==3.1.2
 kiwisolver==1.3.2
 markdown-it-py==2.2.0
 MarkupSafe==2.1.2
-matplotlib==3.4.3
+matplotlib==3.8.3
 mdit-py-plugins==0.3.5
 mdurl==0.1.2
 myst-parser==1.0.0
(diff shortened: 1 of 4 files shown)


bsweger:
These changes feel like they might not be enough information for people to decide if they should use cloud storage (for example, how does it work?)

Would be grateful for suggestions on a useful level of detail and where we might put additional info. Or perhaps it makes sense to start with this minimal information and iterate based on feedback?

micokoch:
I vote to merge it and iterate as needed. If there are improvements, they can be documented as they happen.

bsweger:
@micokoch Makes sense--better to get something out there that reflects the recent `admin.json` schema change!

Taking your advice, will remove the PR's draft label 😄

micokoch (approved):
Looks good to me.

annakrystalli (approved):
I agree @bsweger that more details on what's on offer is eventually required, ideally in a new page that we can add to as the cloud infrastructure is further developed. But for now I think this is fine to merge. Perhaps opening an issue where we can sketch out what should be in the new cloud page might be a good idea.

Issue #24, opened 2022-11-30
Hub-metadata file


harryhoch:
The directory structure specifies that admin.json and tasks.json should be JSON files, but model-metadata-schema.json can be a JSON or yaml file. 

Clarification is needed here.  At the very least, change to say "model-metadata-schema.{json,yml} is a JSON or YAML file". More generally, can we allow both formats for all 3 config files? If not, we should explain why?

mzorn-58:
Will admin.json and tasks.json be able to be a YAML file?  If not, why not? [https://hubdocs.readthedocs.io/en/latest/format/hub-config.html](url)

Issue #99, opened 2024-03-12
Issue on page /quickstart-hub-admin/tasks-config.html


nickreich:
In section 5.5.1 the text about how to designate a output_type as required or optional is not fully clear for the mean output type. I think that if the mean is required, the block should look like
```
"mean": {
                            "output_type_id": {
                                "required": ["NA"],
                                "optional": null
                            }, ...
```
where `["NA"]` is kind of like an indicator that that option has been selected. But the text needs to be clearer.

nickreich:
Noting that there is a more complete description in the schema file, but it would be useful to have here as well.
https://raw.githubusercontent.com/Infectious-Disease-Modeling-Hubs/schemas/main/v2.0.1/tasks-schema.json

micokoch:
Addressed in #102 .

PR #102, opened 2024-03-27
Explaining mean in tasks.json

micokoch:
Adding explanations to address #99 .


micokoch:
Resolves #24 and #99 .

docs/source/quickstart-hub-admin/tasks-config.md
```diff
@@ -98,11 +98,11 @@ Now, read below for details on some of the lines of code in this file:
 ```
 
 ### 5.5. Defining [`"output_type"`](#model_output_format):  
-- The [`output_type`](#model_output_format) is used to establish the valid model output types for a given modeling task. In this example they include `mean` and `quantile`.  
+- The [`output_type`](#model_output_format) is used to establish the valid model output types for a given modeling task. In this example they include `mean` and `quantile`, but `median`, `cdf`, `pmf`, and `sample` are other supported output types. Output types have two additional properties, an `output_type_id` and  a `value` property, both of which establish the valid values that can be entered for this output type.
```
nickreich, line 101: Not a request for a change here, but flagging that based on ongoing work on sample output types, we might eventually change this so that there would be an `output_type_id_params` argument in there as well.

PR #109, opened 2024-04-08
Update software.md

mmkerr:
Added text to indicate that all individual software packages are R packages

### DIFF
--- a/docs/source/user-guide/software.md
+++ b/docs/source/user-guide/software.md
@@ -4,15 +4,15 @@ To assist users in building a hub, we have developed a software suite with speci
 
 ## [`hubData`](https://github.com/Infectious-Disease-Modeling-Hubs/hubData)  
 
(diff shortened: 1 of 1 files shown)


LucieContamin (commented):
Looks good! Thank you!
Should we add hubExamples in the list?
Maybe something like:

> `hubExamples` is an R package that provides example data for forecasting and scenario modeling hubs in the hubverse format. You can find the complete package and documentation [here](https://github.com/Infectious-Disease-Modeling-Hubs/hubExamples).  

And should we order the list alphabetically?

mmkerr:
@LucieContamin Sounds good! I'll put the packages in alphabetical order and add hubExamples

mmkerr:
@LucieContamin Ok, packages are now in alphabetical order and hubExamples has been added. Thanks!!

Issue #98, opened 2024-03-08
Add documentation of target metadata


elray1:
Need some content in this section: https://hubdocs.readthedocs.io/en/latest/user-guide/tasks.html#target-metadata

elray1:
It would be good to include something like a table of the possible values for `target_type` and the `output_type`s with which they could be used.

PR #111, opened 2024-04-12
Add target metadata table 

mmkerr:
@elray1 In response to issue #98 , spoke with @nickreich and added a target metadata table, which also incorporates information from Zoltar (https://docs.zoltardata.com/targets/#valid-prediction-types-by-target-type).

### DIFF
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -84,7 +84,7 @@ It is composed of the following fields:
 * `time_unit`: the units of the time steps in terms of day, week, or month.
 
 ### Example
-Here is an example of how the target metadata fields might appear in the `tasks.json schema` for a Hub whose target is incident covid hospitalizations. 
+Here is an example of how the target metadata fields might appear in the `tasks.json` schema for a Hub whose target is incident covid hospitalizations. 
 
 ```
  "target_metadata": [
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -68,14 +68,14 @@ It is composed of the following fields:
 * `target_units`: the unit of observation used for this target. 
 * `target_keys`: a set of one or more name/value pairs that must match a target defined in the `task_ids` section of the schema. Each value, or the combination of values if multiple keys are specified, defines a single target value.
 * `description`: a verbose explanation of the target, which might include details on the measure used for the target or a definition of 'rate', for example. 
-* `target_type`: the target’s statistical data type. The following table lists the possible values for `target_type` and the `output_type` with which they can be used. 
+* `target_type`: the target’s statistical data type. The following table lists the possible values for `target_type` and the `output_type` with which they can be used. We note that for the binary data type row, mean and median `output_type` are X'ed for definitional consistency, but in practice the hubverse recommends using pmf or sample `output_type` as a more natural way to represent these values.
 
 | `target_type` | mean | median | quantile | cdf   | pmf   | sample 
 |--------- | ----------- |----------- | ----------- |----------- |----------- |----------- |
 | continous | X | X | X | X | - | X |
 | discrete | X | X | X | X | X | X |
 | nominal | - | - | - | - | X | X |
-| binary | - | X | - | - | X | X |
+| binary | X | X | - | - | X | X |
 | date | X | X | X | X | X | X |
 | ordinal | - | X | X | X | X | X |
 | compositional | X | X | - | - | - | X |
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -84,14 +84,22 @@ It is composed of the following fields:
 * `time_unit`: the units of the time steps in terms of day, week, or month.
 
 ### Example
-Here is an example of a Hub that uses incident covid hospitalizations as a target. 
-| field |	value
-|--------- | ----------- |
-| `target_id` |	inc covid hosp
-| `target_name` |	Daily incident COVID hospitalizations
-| `target_units` |	count
-| `target_keys`	| inc covid hosp
-| `description` |	Daily newly reported hospitalizations where the patient has COVID, as reported by hospital facilities and aggregated in the HHS Protect data collection system
-| `target_type` |	discrete
-| `is_step_ahead` |	true
-| `time_unit` |	day
+Here is an example of how the target metadata fields might appear in the `tasks.json schema` for a Hub whose target is incident covid hospitalizations. 
+
+```
+ "target_metadata": [
+                    {
+                       "target_id": "inc covid hosp",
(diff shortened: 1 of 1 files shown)


nickreich:
@elray1 would be helpful to have you double-check our first pass on which output types are valid for each target type.

nickreich (changes requested):
two main comments here:

1. We should make all the items in the table row and column headings lower case. e.g. "Continuous" --> "continuous".
2. Can we add more text here that (a) describes the purpose of the target metadata section of the schema file and (b) defines what each field in the target metadata means? E.g. the target_type field is just one element of the target metadata, so this table should be just one part of this new section.

elray1:
Here are some questions about the alignment between output types and target types:

- There is an X under pmf for the continuous type.  This only makes sense if a continuous variable (hospital admission per 100k pop) has been discretized or binned though, right?  Maybe we should take that one out?
- Binary: I would say a mean is appropriate for a binary target type.  I'm iffy on whether a median is very meaningful; I guess it would just be the most likely category.
    - i initially had a question about whether cdf should be allowed for binary type, but i ended up agreeing with you. If cdf outputs were collected, they would all be 1 for the "yes" category.  Not very informative or worth collecting.

I agreed with everything else in the table.

nickreich:
- I agree with Evan about removing the X in the [continuous, pmf] cell.
- I think we should not make any other changes. agree that median is not that meaningful for binary, but also think it's ok to leave as is.

mmkerr:
@elray1 @nickreich I don't know if I need to do more to address Evan's last comment, but for now I:
-removed the X in the continuous pmf cell
-added text that describes target metadata and defines each field
-added an example (can delete if this is not helpful, just noticed that the output type section above had one)

docs/source/user-guide/tasks.md
```diff
@@ -60,4 +60,38 @@ The [output_type](output_types) object defines accepted representations for each
...
 ## Target metadata
 
-Document here the properties of a target, as listed in the schema.
+Target metadata is an array of objects that defines the characteristics of each target.
+
+It is composed of the following fields:
+* `target_id`: a short description that uniquely identifies the target.
+* `target_name`: a longer, human readable description of the target, which could be used as a visualization axis label.
+* `target_units`: the unit of observation used for this target. 
+* `target_keys`: a value that must match a target set in `task_ids`, to appropriately identify it. Each property should have one specified value. Each value, or the combination of values if multiple keys are specified, defines a single target value. The value should be null in the case where the target is not specified as a `task_id` and is specified solely through the `target_id` `target_metadata` property.
```
nickreich, line 69: I think we should be more specific here because sometimes this can be just a single name/value pair like
```
{ "target": "peak week hosp" }
```
but in more complex cases the target_keys field can be defined by multiple name/value pairs, like
```
{ 
    "target": "peak hosp" , 
    "target_variable": "case"
}
```

So, removing some words that I think make it more complicated to understand, maybe we could say

* `target_keys`: a set of one or more name/value pairs that must match a target defined in the `task_ids` section of the schema. Each value, or the combination of values if multiple keys are specified, defines a single target value.

docs/source/user-guide/tasks.md
```diff
@@ -60,4 +60,38 @@ The [output_type](output_types) object defines accepted representations for each
...
+| nominal | - | - | - | - | X | X |
+| binary | - | X | - | - | X | X |
+| date | X | X | X | X | X | X |
+| ordinal | - | X | X | X | X | X |
+| compositional | X | X | - | - | - | X |
+
+* `is_step_ahead`: a Boolean value that indicates whether the target is part of a sequence of values. If `is_step_ahead` is true, then this field is required and defines the unit of time steps. If `is_step_ahead` is false, then this should be left out and/or will be ignored if present.
+* `time_unit`: the units of the time steps in terms of day, week, or month.
+
+### Example
```
nickreich, line 86: I don't feel that strongly about this, but I think we could consider including this example as actual JSON, which is how it would have to appear in the schema, rather than as a table. So the relevant portion of a specific tasks.json file could be copied and pasted here instead of the table.

docs/source/user-guide/tasks.md
```diff
@@ -60,4 +60,38 @@ The [output_type](output_types) object defines accepted representations for each
 (target_metadata)=
 ## Target metadata
 
-Document here the properties of a target, as listed in the schema.
+Target metadata is an array of objects that defines the characteristics of each target.
```
nickreich, line 63: Suggest adding text so that it reads "Target metadata is an array in the tasks.json schema file that defines the characteristics of each target."

nickreich (changes requested):
This reads much more clearly with the additional detail and context. I have a few suggested comments.

docs/source/user-guide/tasks.md
```diff
@@ -60,4 +60,46 @@ The [output_type](output_types) object defines accepted representations for each
...
+* `target_keys`: a set of one or more name/value pairs that must match a target defined in the `task_ids` section of the schema. Each value, or the combination of values if multiple keys are specified, defines a single target value.
+* `description`: a verbose explanation of the target, which might include details on the measure used for the target or a definition of 'rate', for example. 
+* `target_type`: the target’s statistical data type. The following table lists the possible values for `target_type` and the `output_type` with which they can be used. 
+
+| `target_type` | mean | median | quantile | cdf   | pmf   | sample 
+|--------- | ----------- |----------- | ----------- |----------- |----------- |----------- |
+| continous | X | X | X | X | - | X |
+| discrete | X | X | X | X | X | X |
+| nominal | - | - | - | - | X | X |
+| binary | - | X | - | - | X | X |
```
nickreich, line 78: Noting one additional request here, pulled from @elray1's suggestion [here](https://github.com/Infectious-Disease-Modeling-Hubs/hubDocs/pull/111#issuecomment-2060090238). He says:
> To me, it seems like if we're X-ing things that are definitionally consistent, we should put an X under both mean and median for binary, and if we're X-ing things that hubverse recommends I don't think I'd personally put an X under either (people talk about the mean of binary variables all the time, but in a hub context a setup using pmf feels more natural; and I've just never heard of anyone talk about the median of a binary variable before)

So my suggestion is to add an X in the [binary, mean] cell and to revise the bullet text for taerget_type to be:
> the target’s statistical data type. The following table lists the possible values for `target_type` and the `output_type` with which they can be used. We note that for the binary data type row, mean and median `output_type` are X'ed for definitional consistency, but in practice the hubverse recommends using pmf or sample `output_type` as a more natural way to represent these values.

nickreich (changes requested):
one additional requested change, to resolve comment from Evan.

docs/source/user-guide/tasks.md
mmkerr: @nickreich Done!

Issue #52, opened 2023-06-12
Source info from slides to provide higher level background to docs


annakrystalli:
Inspired by the slides from the recent presentation, it could be really nice to add some of the information in the slides as additional background in our docs.

In particular I feel the slide sections on:

-  overall motivation behind hubs
- their evolution 
- info on types of hubs (i.e. distinction between forecasting, nowcasting, scenarios etc)

would add useful context to the project in our docs.

nickreich:
Here is a link to the "published" slides that we could incorporate. https://docs.google.com/presentation/d/e/2PACX-1vRhXa7DgUDATi3Ovzc8rmUADD9aWeaZKkDV-2wXmja4KT0PJV7elsWiNRHCj9ypRaFEGoFNumOP2mhS/pub?start=false&loop=false&delayms=60000&slide=id.p

mzorn-58:
Figure out what to include from [slides](https://docs.google.com/presentation/d/e/2PACX-1vRhXa7DgUDATi3Ovzc8rmUADD9aWeaZKkDV-2wXmja4KT0PJV7elsWiNRHCj9ypRaFEGoFNumOP2mhS/pub?start=false&loop=false&delayms=3000&slide=id.g20a596752b5_0_151)

micokoch:
Added selected information and slides on PR #112 . If PR is approved, will close this issue.

PR #112, opened 2024-04-16
Adding some history and motivation to hubDocs

micokoch:
Adding info from Hubverse intro presentation to hubDocs - issue #52 in order to provide higher level background information.


nickreich (changes requested):
Thanks for this detailed PR that adds a ton of useful information! I have a few minor requests:

- [x] the image added under key definitions feels a little bit out of place. I like the idea of adding it here, but maybe it should be under its own header (e.g., it's not really a key definition), so that the list of key definitions doesn't just abruptly follow the figure with no explanation.
- [ ] Could you add a citation to the AJPH paper on the history page, which also catalogues some of these same points, and serves as sort of the source of the history of hubs figure (although the figure in that paper is a bit of an older version): https://ajph.aphapublications.org/doi/full/10.2105/AJPH.2022.306831

micokoch:
Hi @nickreich . I addressed all your comments/requests, except adding the `json` tag, for reasons explained below your comment. Let me know if you want me to add it anyway and if you need any additional changes. Thanks!

nickreich (approved):
Thanks, @micokoch ! LGTM.

Issue #115, opened 2024-04-26
Description of model output columns is unclear


lshandross:
In the [Formats of model output](https://github.com/Infectious-Disease-Modeling-Hubs/hubDocs/blob/main/docs/source/user-guide/model-output.md#formats-of-model-output) subsection of the model-output page, the way that the different columns are explained seem to differ between the first paragraph and the bullet points below, which can lead to confusion.

In the paragraph, the documentation says that "columns define: (1) the model task, (2) specification of the representation of the model output, and (3) the model output value." This explanation separates the `value` column as distinct from the `output_type` and `output_type_id` columns. However, in the bullet points below, all three of these columns are described as part of the "model output representation" category.

We should choose a single way of defining the groupings, i.e. deciding whether or not the `value` column is considered part of the "model output representation" group, and update the documentation for consistency to avoid confusing users.

elray1:
Good suggestion, Li, thanks.  My personal preference is to describe `value` separately from `output_type` and `output_type_id`, because the `value` is provided by the modeler while the others are specified by the hub.

micokoch:
@lshandross - I used the hubEnsembles manuscript verbiage to clarify the issue you brought up. I used much of your own wording (I hope that's okay), because it seemed better than me paraphrasing an already clear explanation.
@elray1 - I assumed that, since it's in the manuscript, it had been agreed that the `output_type`, `output_type_id`, and `value` are all part of the "model output representation. If this is incorrect, then, it should probably be changed in the manuscript as well.
I put all the changes in PR #117. If approved, then I will close this issue. Thanks for taking a look at this.

lshandross:
@micokoch I had written the content in the manuscript based on the bullet points in the documentation and hadn't realized until several weeks ago that the documentation was not consistent, so I don't feel like the `output_type`, `output_type_id`, and `value` columns have definitively been decided to be part of the "model output representation." Thus, I feel like we should make a conclusive decision one way or another, then make changes in the appropriate places based on that decision.

PR #117, opened 2024-05-15
Clarify model output conventions

micokoch:


micokoch:
Addresses issue #115 .

micokoch:
@lshandross @elray1 @nickreich - Here is the PR I mentioned, and I added the reference to the hubEnsembles manuscript. Hope that works for you all.

nickreich (changes requested):
Thanks for this PR! I'd suggest not including the journal name, as it hasn't been accepted yet. Otherwise, LGTM.

lshandross (commented):
If we decide that this method of explaining model output conventions where we include the `value` column in the "model output representation columns" (as opposed to `value` being its own type of column), I'm fine with this wording. More discussion on the original issue, Issue #115

PR #125, opened 2024-06-06
Add note about accepted compression ext prefixes

annakrystalli:
File names are now allowed to contain the following compression extension prefixes: .snappy, .gzip, .gz, .brotli, .zstd, .lz4, .lzo, .bz2. This PR adds a note about that to make users/hubAdmins aware


LucieContamin (approved):
Looks good to me, thanks!

PR #136, opened 2024-06-25
Adding Statcounter

micokoch:
Hi @nickreich - I built it locally, and I think I added the counter correctly, but I'm not really sure. I'm not sure how to test that the script is working as intended.


nickreich (approved):
I checked the statcounter site and could see hits from the netlify preview site, which I think indicates it is working. I will approve, merge and monitor the stat counter site stats.

PR #137, opened 2024-06-26
Remove June 17 caution notice and update example_workflow (hubverse-org)

micokoch:


nickreich (commented):
@micokoch The change to remove the caution notice is straightforward. 

Can you explain the example_workflow changes? where were these visible earlier in the documentation and why are they being removed now?

nickreich (approved):
Thanks for the changes!

PR #146, opened 2024-07-15
fix checklist formatting in quickstart

zkamvar:


nickreich (approved):
Y'know - I noticed the no-space in those brackets when we were talking earlier. Thank you for fixing this! I checked the netlify preview and it looks much better!

PR #147, opened 2024-07-16
enable tasklists

zkamvar:
The MyST parser does not include task lists by default, so the extension has to be enabled.


nickreich:
Noting that this seems to add explicit check boxes when using `[ ]` in markdown, as on https://hubverse.io/en/latest/quickstart-hub-admin/intro.html

PR #156, opened 2024-07-24
Adding packages, changing order, others

micokoch:
These are the main changes made to the software section with explanations:

1. `hubverse` is listed as the first package, since it includes all the other packages, and may be the only one needed to download. I also bolded that it includes the rest of the packages.
2. I added descriptions for packages that were not listed in this section (and are part of `hubverse` - namely `hubCI` and `hubEvals`.
3. After playing with the order of packages, and going through their descriptions and functions, there was no obvious hierarchy I could see in which to list them, and opted to list them in alphabetical order. Given there are many packages now, this provides two main benefits: (i) it is easier to find a specific package if they are in order; and (ii) it aligns more closely with how they would display if you only download `hubverse`.
4. I changed the hyperlinked text from "here" to "the complete package and instructions here". This is for accessibility issues for people who use screen readers. Here is [one source on accessible hyperlinks](https://www.nysed.gov/webaccess/create-accessible-hypertext-links).

### DIFF
--- a/docs/source/user-guide/software.md
+++ b/docs/source/user-guide/software.md
@@ -12,7 +12,7 @@ To assist users in building a hub, we have developed a software suite with speci
 
 ## [`hubCI`](https://github.com/hubverse-org/hubCI)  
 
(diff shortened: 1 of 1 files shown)


mmkerr (approved):
Looks great. Agree with all the changes.

PR #162, opened 2024-07-31
change hubdocs URL to hubverse.io

zkamvar:


nickreich:
I am confused about why this PR that is about changing links has a change to a PDF file of old slides. Maybe we should just not change anything here.

zkamvar:
Ah sorry about that. This was the result of a brute-force throwaway script that used this command to do the dirty work:

```sh
sed -i -r -e 's/hubdocs.readthedocs/hubverse/g' $(find . -type f -not -path '*/.git/*')
```

PR #165, opened 2024-08-02
add list of active hubs on landing page

### LINKED ISSUE #149: Add list of hubverse hubs to documentation site
There are many ways to do this, but the content from [this doc](https://docs.google.com/document/d/1jVb58Cert1ON-Do1TdwBxpeiFHMkjqDLofgV2oqezOo/edit) needs to be added to the website. In particular, just the "active hubs using hubverse" should be added. Ideally there would be a list with some standard way of representing the data. Perhaps just the "name" of the hub, a brief description and a link to the repo and/or website for each one?

My suggestion might be to make this its own page under the "Overview" section, but I don't feel strongly about that. I'm not sure what the title should be "List of hubs"?

zkamvar:
I added hubs from the list of Active and planned hubs. Of the ones present in the document (linked from https://github.com/hubverse-org/hubDocs/issues/149#issue-2411680480). I'm happy to change the representation and add/remove information as needed. 

To display the hubs, I used the `sphinx-design` module, which gives us the ability to display cards. 

I only added hubs that were public, active, and had links to the content.

This will fix #149


micokoch (approved):
It looks great, and I like the design, but @nickreich should probably look at what hubs to include on this list.

zkamvar:
Agreed. I've added him as a reviewer.

nickreich (changes requested):
I really really like the idea of having these "cards" on the landing page, as I think they provide some very valuable context (and implicit success stories) for folks new to the hubverse.

But, I think we should standardize how we present some of the content here. For example, maybe we have a main link that is to the public facing site of the challenge (if there is one). [E.g. flusight has this page https://www.cdc.gov/flu-forecasting/about/index.html, which is kind of like the "website" link for the respicast page.]

So my suggestion would be to have these fields if they exist:
 - repo
 - website

And then have a set of links that are secondary (maybe labeled "resources"?) that include things like visualization, evaluation reports, etc...

Other minor comments: I'd remove the named contact for the WNV repo and use linked words/phrases rather than including the entire link url address as the link.

nickreich:
It appears that the RTD preview hasn't updated to reflect the new changes. Since the appearance of the "cards" is important here, can we trigger that somehow?

zkamvar:
> It appears that the RTD preview hasn't updated to reflect the new changes. Since the appearance of the "cards" is important here, can we trigger that somehow?

It should have. 
https://hubdocs--165.org.readthedocs.build/en/165/#active-hubs

The changes look like this:

screencapture of a section showing four active hubs in a grid of cards

nickreich (changes requested):
Sorry - I just had to refresh the page to get the new updates to load. I'd like to be a stickler about a few minor changes here:

 - I can't find any public resources about the WNV forecasts on the CA DPH website, so I'd maybe suggest removing that link. This is different from FluSight, where there is a dedicated page to the FluSight project.
 - I liked the previous set up for the github repo links: github icon and the text of the link being the `username/reponame` structure. Could we move back to that for the repo links only? Maybe something like 
     - Repo: [github icon] username/reponame
 - remove "v3" from the FluSight header and make it (2023-2025)

nickreich (approved):
Thanks for bearing with all my small change requests. Looks great!

PR #196, opened 2024-10-21
Correct modelling section to refer to hubValidations as useful package

annakrystalli:
Removed reference to `hubAdmin` and added more detail about useful function to submitting modellers


zkamvar (approved):
Thank you! The exact function names were escaping me when I wrote this at first.

PR #197, opened 2024-10-21
clarify missingness of point estimates; add examples for determining output_type_id format

zkamvar:
This modifies the description of `output_type_id` for point estimates to clearly state that the entry should be a missing value, either `NA` in R or `None` in Python. 

This also states that this applies to schemas prior to v4.0.0 (see https://github.com/hubverse-org/schemas/issues/109) and adds the python one-liner as suggested by @MKupperman in https://github.com/hubverse-org/hubDocs/issues/198.

The preview for this PR is at https://hubdocs--197.org.readthedocs.build/en/197/user-guide/model-output.html

### DIFF
--- a/docs/source/user-guide/model-output.md
+++ b/docs/source/user-guide/model-output.md
@@ -46,7 +46,9 @@ one-week-ahead incidence, but probabilities for the timing of a season peak:
 | EW202242 | weekly rate | 1 | sample | 2 | 3 |
 :::
 
-[^batman]: `NA` (without quotes) indicates missingness in R, which is the expected `output_type_id` for a `mean` `output_type`. 
+[^batman]: The `output_type_id` for point estimates (e.g. `mean`) is not applicable. To
+reflect this, we need to signal that this is a missing value. In R, missing values are
+encoded as `NA`, and in Python, they are encoded as `None`.
   This is discussed in the [output type table](#output-type-table)
 
 
--- a/docs/source/user-guide/model-output.md
+++ b/docs/source/user-guide/model-output.md
@@ -31,7 +31,7 @@ one-week-ahead incidence, but probabilities for the timing of a season peak:
 :::{table} An example of a model output submission for modelA
 | `origin_epiweek` | `target` | `horizon` | `output_type` | `output_type_id` | `value` |
 | ------ | ------ | ------ | ------ | ------ | ------ | 
-| EW202242 | weekly rate | 1 | mean     | NA | 5 |
+| EW202242 | weekly rate | 1 | mean     | NA[^batman] | 5 |
 | EW202242 | weekly rate | 1 | quantile | 0.25 | 2 |
 | EW202242 | weekly rate | 1 | quantile | 0.5 | 3 |
 | EW202242 | weekly rate | 1 | quantile | 0.75 | 10 |
@@ -45,6 +45,10 @@ one-week-ahead incidence, but probabilities for the timing of a season peak:
 | EW202242 | weekly rate | 1 | sample | 1 | 3 |
 | EW202242 | weekly rate | 1 | sample | 2 | 3 |
 :::
+
+[^batman]: `NA` (without quotes) indicates missingness in R, which is the expected `output_type_id` for a `mean` `output_type`. 
+  This is discussed in the [Output type table](#output-type-table)
+
 
 (formats-of-model-output)=
 ### File formats
@@ -112,8 +116,8 @@ The following table provides more detail on how to configure the three "model ou
 :::{table} Relationship between the three model output representation columns with respect to the type of prediction (`output_type`)
 | `output_type` | `output_type_id` | `value` |
 | ------ | ------ | ------ | 
-| `mean` | `"NA"`[^batman] (not used for mean predictions) | Numeric: the mean of the predictive distribution |
-| `median` | `"NA"` (not used for median predictions) | Numeric: the median of the predictive distribution |
+| `mean` | None[^missingno] (not used for mean predictions) | Numeric: the mean of the predictive distribution |
+| `median` | None (not used for median predictions) | Numeric: the median of the predictive distribution |
 | `quantile` | Numeric between 0.0 and 1.0: a probability level | Numeric: the quantile of the predictive distribution at the probability level specified by the output_type_id |
 | `cdf`[^cdf] | String or numeric: a possible value of the target variable | Numeric between 0.0 and 1.0: the value of the cumulative distribution function of the predictive distribution at the value of the outcome variable specified by the output_type_id |
 | `pmf`[^pmf] | String naming a possible category of a discrete outcome variable | Numeric between 0.0 and 1.0: the value of the probability mass function of the predictive distribution when evaluated at a specified level of a categorical outcome variable.[^cdf] |
@@ -121,10 +125,11 @@ The following table provides more detail on how to configure the three "model ou
 :::
 
 
-[^batman]: Why have `"NA"` as the `output_type_id`? There are two reasons for this. 
-  First, this provides a placeholder for the model output CSV file in the presence of other output types for validation.
-  The second reason is that we already use `null` to indicate the presence of an absence in the `required` and `optional` fields, and having this allows hubverse tools to treat point estimates without special handling. 
-
+[^missingno]: Point estimates don't have an `output_type_id` because you can only have one point estimate for each combination of task IDs.
+   However, because the `output_type_id` column is requrired, something has to go in this place, which is a missing value. 
+   This is encoded as [`NA_character_` in R](https://www.njtierney.com/post/2020/09/17/missing-flavour/) (which is why our schemas prior to 4.0.0 encoded these as `["NA"]`).
+   **If you use Python** to write parquet files, the `output_type_id` column should be an array of `None` and you will need to [explicitly cast the `output_type_id` column as a "string"](https://github.com/hubverse-org/hubValidations/issues/131#issuecomment-2427654006) (via Pandas):    
+   `df["output_type_id"] = df["output_type_id"].astype("string")`
 [^pmf]: **Note on `pmf` model output type**: Values are required to sum to 1 across all `output_type_id` values within each combination of values of task ID variables. This representation should only be used if the outcome variable is truly discrete; a CDF representation is preferred if the categories represent a binned discretization of an underlying continuous variable.
 
 [^sample]: **Note on `sample` model output type**: Depending on the hub specification, samples with the same sample index (specified by the `output_type_id`) may be assumed to correspond to a single sample from a joint distribution across multiple levels of the task ID variables — further details are discussed below.
--- a/docs/source/user-guide/model-output.md
+++ b/docs/source/user-guide/model-output.md
@@ -101,12 +101,12 @@ specified in the [section on usage of task ID variables](#task-id-use) when appr
 2. **"Model output representation" (3 columns)**: consists of three
    columns specifying how the model outputs are represented. All three of these
    columns will be present in all model output data:
-  1. `output_type` specifies the type of representation of the predictive
-     distribution, namely `"mean"`, `"median"`, `"quantile"`, `"cdf"`, `"cmf"`,
-     `"pmf"`, or `"sample"`.
-  2. `output_type_id` specifies more identifying information specific to the
-     output type, which varies depending on the `output_type`.
-  3. `value` contains the model’s prediction.
+   1. `output_type`{.codeitem} specifies the type of representation of the predictive
+      distribution, namely `"mean"`, `"median"`, `"quantile"`, `"cdf"`, `"cmf"`,
+      `"pmf"`, or `"sample"`.
+   2. `output_type_id`{.codeitem} specifies more identifying information specific to the
+      output type, which varies depending on the `output_type`.
+   3. `value`{.codeitem} contains the model’s prediction.
 
 
 The following table provides more detail on how to configure the three "model output representation" columns based on each model output type.
@@ -123,7 +123,7 @@ The following table provides more detail on how to configure the three "model ou
 | `sample` | Positive integer sample index | Numeric: a sample from the predictive distribution.
 :::
 
-:::{note} Caveats for model `output_type_id`s
+:::{note} 
 :name: output-type-caveats
 
 The model output type ids have different caveats depending on the `output_type`:
--- a/docs/source/user-guide/model-output.md
+++ b/docs/source/user-guide/model-output.md
@@ -92,72 +92,12 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
 
 [^model-id]: When using models for downstream analysis with the [`collect_hub()` function](https://hubverse-org.github.io/hubData/reference/collect_hub.html) in the `hubData` package, one more column called `model_id` is prepended that identifies the model from its filename. 
 
-The model output follows the specification of the `tasks.json` configuration
-file of the hub. If you are creating a model and would like to know what
-data type your columns should be in, the hubVerse has utilities to provide an
-arrow schema and even a full submission template from the `tasks.json`
-configuration file. 
-
-### Arrow Schema
-
-**The hubverse packages `hubData()` and `hubUtils()` have functionality that will generate an
-arrow schema so that you can ensure your output matches the expected type.** 
-
-Here is some example code that can help. In this example, `tasks_path` is the
-path to your `tasks.json` config file (note: it can also be a URL like [the
-simple forecasts example hub `tasks.json`
-file](https://raw.githubusercontent.com/hubverse-org/example-simple-forecast-hub/refs/heads/main/hub-config/tasks.json))
-
-```r
-# read the configuration file and get the latest round
-config_tasks <- hubUtils::read_config_file(tasks_path)
-schema <- hubData::create_hub_schema(config_tasks)
-```
-
-The schema output will look something like this:
-
-```
-Schema
-origin_date: date32[day]
-target: string
-horizon: int32
-location: string
-age_group: string
-output_type: string
-output_type_id: double
-value: int32
-model_id: string
-```
-
-### Submission Template
-
-**The hubverse package `hubValidations()` has functionality
-that will generate template data to get your started.** This submission
-template can be written as a CSV or parquet file and then imported in to
-whatever software you use to run your model.
-
-Here is some example code that can help. In this example, `tasks_path` is the
-path to your `tasks.json` config file (note: it can also be a URL like [the
-simple forecasts example hub `tasks.json`
-file](https://raw.githubusercontent.com/hubverse-org/example-simple-forecast-hub/refs/heads/main/hub-config/tasks.json))
-
-```r
-# read the configuration file and get the latest round
-config_tasks <- hubUtils::read_config_file(tasks_path)
-rounds <- hubUtils::get_round_ids(config_tasks)
-this_round <- rounds[length(rounds)]
-# create the submission template (this may take some time if your submission uses samples)
-tmpl <- hubValidations::submission_tmpl(config_tasks = config_tasks, round_id = this_round)
-# write the template to a parquet file to use in your model code. 
-arrow::write_parquet(tmpl, "/path/to/template.parquet")
-```
 
 (column-details)=
 ### Details about model output column specifications
 
 As shown in the [model output submission table](#model-output-example-table) above, there are three **"task ID"** columns: `origin_epiweek`, `target`, and `horizon`; and there are two **"model output representation"** columns: `output_type` and `output_type_id` followed by the `value` column.  
 More detail about each of these column groups is given in the following points:  
-
 
 1. **"Task IDs" (multiple columns)**:  The details of the outcome (the model task) are provided by the modeler and can be stored in a series of "task ID" columns as described in this [section on task ID variables](#task-id-vars). These "task ID" columns may also include additional information, such as any conditions or assumptions used to generate the predictions. Some example task ID variables include `target`, `location`, `reference_date`, and `horizon`. Although there are no restrictions on naming task ID variables, we suggest that hubs adopt the standard task ID or column names and definitions specified in the [section on usage of task ID variables](#task-id-use) when appropriate.  
 2. **"Model output representation" (2 columns)**: consists of two columns specifying how the model outputs are represented. Both of these columns will be present in all model output data:  
@@ -213,11 +153,15 @@ variables — further details are discussed below.
 
 :::
 
-## 
-
 
 (writing-model-output)=
 ## Writing model output to a hub
+
+The model output follows the specification of the `tasks.json` configuration
+file of the hub. If you are creating a model and would like to know what
+data type your columns should be in, the hubVerse has utilities to provide an
+arrow schema and even a full submission template from the `tasks.json`
+configuration file. 
 
 When submitting model output to a hub, it should be placed in a folder with the
 name of your `model_id` in the model outputs folder specified by the hub
@@ -225,17 +169,51 @@ administrator (this is usually called `model-output`). Below are R and Python ex
 for writing Hubverse-compliant model output files in both CSV and parquet format.
 In these examples, we are assuming the following variables already exist:
 
- - `model_out` is the tabular output from your model formatted as specified
-   in [the formats of model output section](#formats-of-model-output).
  - `hub_path` is the path to the hub cloned on your local computer
  - `model_id` is the combination of `<team_abbr>-<model_abbr>` 
  - `file_name` is the file name of your model formatted as
    `<round_id>-<model_id>.csv` (or `.parquet`)
+ - `model_out` is the tabular output from your model formatted as specified
+   in [the formats of model output section](#formats-of-model-output).
+
+### Submission Template
+
+**The hubverse package `hubValidations()` has functionality
+that will generate template data to get your started.** This submission
+template can be written as a CSV or parquet file and then imported in to
+whatever software you use to run your model.
+
+Here is some example code that can help. In this example, `hub_path` is the
+path to the hub on your local machine.
+
+```r
+# read the configuration file and get the latest round
+config_tasks <- hubUtils::read_config(hub_path)
+rounds <- hubUtils::get_round_ids(config_tasks)
+this_round <- rounds[length(rounds)]
+
+# create the submission template (this may take some time if your submission uses samples)
+tmpl <- hubValidations::submission_tmpl(config_tasks = config_tasks, round_id = this_round)
+```
+
+You can then either write this template to a csv file with the `readr` package:
+
+```r
+# write the template to a csv file to use in your model code. 
+readr::write_csv(tmpl, "/path/to/template.csv")
+```
+
+OR you can write it to a parquet file with the `arrow` package:
+
+```r
+# write the template to a parquet file to use in your model code. 
+arrow::write_parquet(tmpl, "/path/to/template.parquet")
+```
 
 (example-csv)=
 ### Example: model output as CSV
 
-The sections below provide examples for writing CSV model output files that correctly encode missing data.
+The sections below provide examples for writing CSV model output files.
 
 #### Writing CSV with R
 
@@ -252,6 +230,7 @@ Python users can use the `pandas` package when creating CSV model output files.
 ```python
 import pandas as pd
 import os.path
+
 # ... generate model data ...
 outfile = os.path.join(hub_path, "model-output", model_id, file_name)
 model_out.to_csv(outfile, index = False)
@@ -260,21 +239,52 @@ model_out.to_csv(outfile, index = False)
 (example-parquet)=
 ### Example: model output as parquet
 
-Unlike a CSV, a parquet files contain embedded information about the data types of its
-columns. Therefore, when writing model output files as parquet, it's critical that you
-first ensure the data type of the `output_type_id` column matches the
-[expected `output_type_id_datatype` property of the schema](#output-type-id-datatype).
+Unlike a CSV, a parquet files contain embedded information about the data types
+of its columns. 
+Therefore, when writing model output files as parquet, it's
+critical that you first ensure the data type of your columns matches the
+expected type from the Arrow schema.
 
 If the data types of the model output parquet file don't match the hub's schema, the
 submission will not validate.
 In practice, you will need to know whether or not the expected data type is a
 **string/character**, **float/numeric**, or an **Int/integer**.
 
+#### Arrow Schema
+
+**The hubverse packages `hubData()` and `hubUtils()` have functionality that will generate an
+arrow schema so that you can ensure your output matches the expected type.** 
+
+Here is some example code that can help. In this example, `hub_path` is the
+path to the hub on your local machine.
+
+```r
+# read the configuration file and get the latest round
+config_tasks <- hubUtils::read_config(hub_path, "tasks")
+schema <- hubData::create_hub_schema(config_tasks)
+```
+
+The schema output will look something like this:
+
+```
+Schema
+origin_date: date32[day]
+target: string
+horizon: int32
+location: string
+age_group: string
+output_type: string
+output_type_id: double
+value: int32
+model_id: string
+```
+
 #### Writing parquet with R
 
 ```r
 # ... generate model data ...
 outfile <- fs::path(hub_path, "model-output", model_id, file_name)
+
 # update the output_type_id data type to match the hub's schema
 model_out$output_type_id <- as.character(model_out$output_type_id) # or as.numeric(), or as.integer()
 arrow::write_parquet(model_out, outfile)
@@ -288,6 +298,7 @@ import pandas as pd
 import os.path
 # ... generate model data ...
 outfile = os.path.join(hub_path, "model-output", model_id, file_name)
+
 # update the output_type_id data type to match the hub's schema
 model_out["output_type_id"] = model_out["output_type_id"].astype("string") # or "float", or "Int64"
 model_out.to_parquet(outfile)
--- a/docs/source/user-guide/model-output.md
+++ b/docs/source/user-guide/model-output.md
@@ -47,7 +47,7 @@ one-week-ahead incidence, but probabilities for the timing of a season peak:
 :::
 
 [^batman]: `NA` (without quotes) indicates missingness in R, which is the expected `output_type_id` for a `mean` `output_type`. 
-  This is discussed in the [Output type table](#output-type-table)
+  This is discussed in the [output type table](#output-type-table)
 
 
 (file-formats)=
@@ -71,7 +71,7 @@ submission formats are _not mutually exclusive_; **hubs may choose between
       * Compatibility: Harder to work with; teams and people who want to work with files need to install additional libraries
 
 Examples of how to create these file formats in R and Python are listed below in
-[the Writing Model Output section](#writing-model-output).
+[the writing model output section](#writing-model-output).
 
 (formats-of-model-output)=
 ## Formats of model output
@@ -129,7 +129,7 @@ The following table provides more detail on how to configure the three "model ou
 :::{note} 
 :name: output-type-caveats
 
-The model output type ids have different caveats depending on the `output_type`:
+The model output type IDs have different caveats depending on the `output_type`:
 
 `mean` and `median`
 : Point estimates do not have an `output_type_id` because you can only have one
@@ -170,7 +170,7 @@ of writing model output to a hub in R and Python using parquet and CSV files.
 In these examples, we are assuming the following variables already exist:
 
  - `model_out_tbl` is the tabular output from your model formatted as specified
-   in [the Formats of Model Output section](#formats-of-model-output).
+   in [the formats of model output section](#formats-of-model-output).
  - `path_to_hub` is the path to the hub cloned on your local computer
  - `model_name` is the file name of your model formatted as
    `<round_id>-<model_name>.csv` (or parquet)
@@ -178,7 +178,7 @@ In these examples, we are assuming the following variables already exist:
 (example-csv)=
 ### Example: model output as CSV
 
-To write to CSV, you would use the `write_csv()` from the readr package in R and
+To write to CSV, you would use the `write_csv()` from the `readr` package in R and
 the `to_csv()` method in Python. 
 
 #### Writing CSV with R
--- a/docs/source/user-guide/model-output.md
+++ b/docs/source/user-guide/model-output.md
@@ -91,7 +91,7 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
 [^model-id]: When using models for downstream analysis with the [`collect_hub()` function](https://hubverse-org.github.io/hubData/reference/collect_hub.html) in the `hubData` package, one more column called `model_id` is prepended that identifies the model from its filename. 
 
 The model output follows the specification of the `tasks.json` configuration
-file of the hub. If you are creating a model and you would like to know what
+file of the hub. If you are creating a model and would like to know what
 data type your columns should be in, the hubVerse has utilities to provide and
 arrow schema and even a full submission template from the `tasks.json`
 configuration file. 
@@ -108,7 +108,7 @@ file](https://raw.githubusercontent.com/hubverse-org/example-simple-forecast-hub
 
 ```r
 # read the configuration file and get the latest round
-config_tasks <- hubData::read_config_file(tasks_path)
+config_tasks <- hubUtils::read_config_file(tasks_path)
 schema <- hubData::create_hub_schema(config_tasks)
 ```
 
@@ -141,11 +141,11 @@ file](https://raw.githubusercontent.com/hubverse-org/example-simple-forecast-hub
 
 ```r
 # read the configuration file and get the latest round
-config_tasks <- hubData::read_config_file(tasks_path)
-rounds <- hubData::get_round_ids(config_tasks)
+config_tasks <- hubUtils::read_config_file(tasks_path)
+rounds <- hubUtils::get_round_ids(config_tasks)
 this_round <- rounds[length(rounds)]
 # create the submission template (this may take some time if your submission uses samples)
-tmpl <- hubValidations::submission_tmpl(config_tasks, round_id = this_round)
+tmpl <- hubValidations::submission_tmpl(config_tasks = config_tasks, round_id = this_round)
 # write the template to a parquet file to use in your model code. 
 arrow::write_parquet(tmpl, "/path/to/template.parquet")
 ```
@@ -219,9 +219,9 @@ variables — further details are discussed below.
 ## Writing model output to a hub
 
 When submitting model output to a hub, it should be placed in a folder with the
-name of your model in the model outputs folder specified by the hub
-administrator (this is usually called `model-output`). Below are two examples
-of writing model output to a hub in R and Python using parquet and CSV files.
+name of your `model_id` in the model outputs folder specified by the hub
+administrator (this is usually called `model-output`). Below are R and Python examples
+for writing Hubverse-compliant model output files in both CSV and parquet format.
 In these examples, we are assuming the following variables already exist:
 
  - `model_out` is the tabular output from your model formatted as specified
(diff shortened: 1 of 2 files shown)


zkamvar:
I would like to get another modeler perspective on this if it sounds clear: @trobacker

docs/source/user-guide/model-output.md
```diff
@@ -112,19 +116,20 @@ The following table provides more detail on how to configure the three "model ou
...
 :::
 
 
-[^batman]: Why have `"NA"` as the `output_type_id`? There are two reasons for this. 
-  First, this provides a placeholder for the model output CSV file in the presence of other output types for validation.
-  The second reason is that we already use `null` to indicate the presence of an absence in the `required` and `optional` fields, and having this allows hubverse tools to treat point estimates without special handling. 
-
+[^missingno]: Point estimates don't have an `output_type_id` because you can only have one point estimate for each combination of task IDs.
+   However, because the `output_type_id` column is requrired, something has to go in this place, which is a missing value. 
+   This is encoded as [`NA_character_` in R](https://www.njtierney.com/post/2020/09/17/missing-flavour/) (which is why our schemas prior to 4.0.0 encoded these as `["NA"]`).
```
annakrystalli, line 130: This is not always encoded as `NA_character_`. Output type ID columns may well be `double` if e.g. only `mean` and `quantile` output type IDs are being collected. The data type will be [determined by the overall config](https://hubverse.io/en/latest/user-guide/model-output.html#the-output-type-id-column-data-type). What we want to make clear is that `output_type_id` columns containing `NA` values must still adhere to the hub schema. Then we can provide tips on how to do this is different languages. E.g. in R you can assign `NA_real_` or  `NA_character_` or coerce to any other data type using the `as.*()` functions.

Also the `"NA"` in the config doesn't really have anything to do with a data type.  It is actually read in as the default logical:
``` r
jsonlite::fromJSON('{"required": ["NA"]}')$required |> typeof()
#> [1] "logical"
```
It's just the way to trick R to read an `NA` in as a bare `NA` is not allowed in JSON

We'd need to do e.g.:
``` r
jsonlite::fromJSON('{"required": ["NA_character_"]}')$required |> typeof()
#> [1] "character"
```
to get character but that's even more R specific so we avoided it.

docs/source/user-guide/model-output.md
```diff
@@ -112,19 +116,20 @@ The following table provides more detail on how to configure the three "model ou
...
+   **If you use Python** to write parquet files, the `output_type_id` column should be an array of `None` and you will need to [explicitly cast the `output_type_id` column as a "string"](https://github.com/hubverse-org/hubValidations/issues/131#issuecomment-2427654006) (via Pandas):
```
annakrystalli, line 131: Not necessarily the entire column but any output type ID values associated with a point estimate output type. This may be the entire column if only point estimate output types are being submitted.

docs/source/user-guide/model-output.md
```diff
@@ -112,19 +116,20 @@ The following table provides more detail on how to configure the three "model ou
 :::{table} Relationship between the three model output representation columns with respect to the type of prediction (`output_type`)
 | `output_type` | `output_type_id` | `value` |
 | ------ | ------ | ------ | 
-| `mean` | `"NA"`[^batman] (not used for mean predictions) | Numeric: the mean of the predictive distribution |
-| `median` | `"NA"` (not used for median predictions) | Numeric: the median of the predictive distribution |
+| `mean` | None[^missingno] (not used for mean predictions) | Numeric: the mean of the predictive distribution |
```
annakrystalli, line 119: I would prefer at the very least something like `NA/None` (without any quotation marks) here with a prompt to more language specific details. That way `NA` is still easily recognisable as a correct value by R users. I believe `<NA>` also exists in python via `pd.NA`, is the closest thing to R's `NA` and I _think_ that's actually what @MKupperman used in his successful exploration (but I might be wrong).

In any case, I've renamed and moved #198 to here and have suggested we test out the various missingness options available in python so we have clarity about what works and what doesn't when we write this part of the documentation. Details in the issue.

docs/source/user-guide/model-output.md
```diff
@@ -112,19 +116,20 @@ The following table provides more detail on how to configure the three "model ou
...
+| `median` | None (not used for median predictions) | Numeric: the median of the predictive distribution |
```
annakrystalli, line 120: Same comment as above

docs/source/user-guide/model-output.md
annakrystalli: Overall I feel this section is important enough and needs enough detailed discussion that it's own section is appropriate rather than a footnote

annakrystalli (changes requested):

docs/source/user-guide/model-output.md
annakrystalli: Also I don't feel a link to an issue comment is the best way to discuss this. As discussed above, I think a full section with more detailed advice would be more appropriate

micokoch (approved):
Anna made the technical comments, but from a readability angle, it looks good to me. It even made me realize I should uncapitalize a "parquet" which I had capitalized previously.

docs/source/user-guide/model-output.md
zkamvar: > Overall I feel this section is important enough and needs enough detailed discussion that it's own section is appropriate rather than a footnote

Yeah I was realizing that nearly every cell in this table had a footnote, which meant that a new section is absolutely needed.

zkamvar:
I've updated the text according to Anna's suggestions and what I found in #198

elray1:
I have not read all of this carefully.  But I noticed Anna's comment above, "I believe <NA> also exists in python via pd.NA, is the closest thing to R's NA and I think that's actually what @MKupperman used in his successful exploration (but I might be wrong)."

Just noting that [missing values in polars](https://docs.pola.rs/user-guide/expressions/missing-data/#null-and-nan-values) are best/most often/most generally represented using `null` (which is in practice set by specifying `None`), but `np.nan` may be used in the case of a column with floating point values.  Again, haven't read this but just wanted to quickly mention it in case it's relevant to anything here.

annakrystalli:
It might indeed be relevant @elray1.

What we are interested here is what missing values will serialise in an interoperable way that will be correctly interepreted as a missing value by our software when read in. We will definitely have `NA`s in our files as we will always have R users writing to `csv` files so either way polars will need to deal with that. I believe it can though as you would expect the R version of the polars package to understand `NA`s. There are also arguments when reading in files that allow us to specify null values on read (in both python and R).

Having said all that, it's of course best not to make any assumptions and actually test to ensure our suggestions don't cause any problems to desired downstream functionality. So @elray1, would you be able to add what a successful test of polars functionality requirements would look like here 🙏? https://github.com/hubverse-org/hubDocs/issues/198

zkamvar:
This PR is intended to clarify that `"NA"` in the schema actually means missing data and not the character `"NA"`. I've tested these assumptions in https://github.com/hubverse-org/hubDocs/issues/198#issuecomment-2430287818 and I think it satisfies this comment: 

> Having said all that, it's of course best not to make any assumptions and actually test to ensure our suggestions don't cause any problems to desired downstream functionality.

docs/source/user-guide/model-output.md
```diff
@@ -46,7 +46,11 @@ one-week-ahead incidence, but probabilities for the timing of a season peak:
 | EW202242 | weekly rate | 1 | sample | 2 | 3 |
 :::
 
-(formats-of-model-output)=
+[^batman]: `NA` (without quotes) indicates missingness in R, which is the expected `output_type_id` for a `mean` `output_type`. 
+  This is discussed in the [Output type table](#output-type-table)
```
micokoch, line 50:
```suggestion
  This is discussed in the [output type table](#output-type-table)
```

docs/source/user-guide/model-output.md
```diff
@@ -66,8 +70,10 @@ submission formats are _not mutually exclusive_; **hubs may choose between
    * Disadvantages:
       * Compatibility: Harder to work with; teams and people who want to work with files need to install additional libraries
 
+Examples of how to create these file formats in R and Python are listed below in
+[the Writing Model Output section](#writing-model-output).
```
micokoch, line 74:
```suggestion
[the writing model output section](#writing-model-output).
```

docs/source/user-guide/model-output.md
```diff
@@ -98,39 +104,135 @@ specified in the [section on usage of task ID variables](#task-id-use) when appr
...
-| `sample`[^sample] | Positive integer sample index | Numeric: a sample from the predictive distribution.
+| `cdf` | String or numeric: a possible value of the target variable | Numeric between 0.0 and 1.0: the value of the cumulative distribution function of the predictive distribution at the value of the outcome variable specified by the output_type_id |
+| `pmf` | String naming a possible category of a discrete outcome variable | Numeric between 0.0 and 1.0: the value of the probability mass function of the predictive distribution when evaluated at a specified level of a categorical outcome variable. |
+| `sample` | Positive integer sample index | Numeric: a sample from the predictive distribution.
+:::
+
+:::{note} 
+:name: output-type-caveats
+
+The model output type ids have different caveats depending on the `output_type`:
```
micokoch, line 132:
```suggestion
The model output type IDs have different caveats depending on the `output_type`:
```

docs/source/user-guide/model-output.md
```diff
@@ -98,39 +104,135 @@ specified in the [section on usage of task ID variables](#task-id-use) when appr
...
+name of your model in the model outputs folder specified by the hub
+administrator (this is usually called `model-output`). Below are two examples
+of writing model output to a hub in R and Python using parquet and CSV files.
+In these examples, we are assuming the following variables already exist:
 
-[^batman]: Why have `"NA"` as the `output_type_id`? There are two reasons for this. 
-  First, this provides a placeholder for the model output CSV file in the presence of other output types for validation.
-  The second reason is that we already use `null` to indicate the presence of an absence in the `required` and `optional` fields, and having this allows hubverse tools to treat point estimates without special handling. 
+ - `model_out_tbl` is the tabular output from your model formatted as specified
+   in [the Formats of Model Output section](#formats-of-model-output).
```
micokoch, line 173:
```suggestion
   in [the formats of model output section](#formats-of-model-output).
```

docs/source/user-guide/model-output.md
```diff
@@ -98,39 +104,135 @@ specified in the [section on usage of task ID variables](#task-id-use) when appr
...
+ - `path_to_hub` is the path to the hub cloned on your local computer
+ - `model_name` is the file name of your model formatted as
+   `<round_id>-<model_name>.csv` (or parquet)
 
-[^pmf]: **Note on `pmf` model output type**: Values are required to sum to 1 across all `output_type_id` values within each combination of values of task ID variables. This representation should only be used if the outcome variable is truly discrete; a CDF representation is preferred if the categories represent a binned discretization of an underlying continuous variable.
+(example-csv)=
+### Example: model output as CSV
 
-[^sample]: **Note on `sample` model output type**: Depending on the hub specification, samples with the same sample index (specified by the `output_type_id`) may be assumed to correspond to a single sample from a joint distribution across multiple levels of the task ID variables — further details are discussed below.
+To write to CSV, you would use the `write_csv()` from the readr package in R and
```
micokoch, line 181:
```suggestion
To write to CSV, you would use the `write_csv()` from the `readr` package in R and
```

micokoch (commented):
It looks very good, as far as I can tell. I made some small capitalization suggestions to be consistent with the rest of hubDocs.

zkamvar:
@annakrystalli, did I address your concerns?

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+When submitting model output to a hub, it should be placed in a folder with the
+name of your model in the model outputs folder specified by the hub
+administrator (this is usually called `model-output`). Below are two examples
+of writing model output to a hub in R and Python using parquet and CSV files.
+In these examples, we are assuming the following variables already exist:
 
-[^pmf]: **Note on `pmf` model output type**: Values are required to sum to 1 across all `output_type_id` values within each combination of values of task ID variables. This representation should only be used if the outcome variable is truly discrete; a CDF representation is preferred if the categories represent a binned discretization of an underlying continuous variable.
+ - `model_out_tbl` is the tabular output from your model formatted as specified
+   in [the formats of model output section](#formats-of-model-output).
+ - `path_to_hub` is the path to the hub cloned on your local computer
```
annakrystalli, line 161:
```suggestion
 - `hub_path` is the path to the hub cloned on your local computer
```
This is the argument name across the hubverse

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+ - `model_name` is the file name of your model formatted as
```
annakrystalli, line 162:
```suggestion
 - `file_name` is the file name of your model formatted as
```

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
-  First, this provides a placeholder for the model output CSV file in the presence of other output types for validation.
-  The second reason is that we already use `null` to indicate the presence of an absence in the `required` and `optional` fields, and having this allows hubverse tools to treat point estimates without special handling. 
+When submitting model output to a hub, it should be placed in a folder with the
+name of your model in the model outputs folder specified by the hub
+administrator (this is usually called `model-output`). Below are two examples
+of writing model output to a hub in R and Python using parquet and CSV files.
+In these examples, we are assuming the following variables already exist:
 
-[^pmf]: **Note on `pmf` model output type**: Values are required to sum to 1 across all `output_type_id` values within each combination of values of task ID variables. This representation should only be used if the outcome variable is truly discrete; a CDF representation is preferred if the categories represent a binned discretization of an underlying continuous variable.
+ - `model_out_tbl` is the tabular output from your model formatted as specified
```
annakrystalli, line 159: This is not quite accurate. Model submissions do not generally contain `model_id` columns, When date is accessed from a hub via `coonect_hub()` it does. Only data accessed by a hub is a `model_out_tbl` class object.

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+   in [the formats of model output section](#formats-of-model-output).
+ - `path_to_hub` is the path to the hub cloned on your local computer
+ - `model_name` is the file name of your model formatted as
+   `<round_id>-<model_name>.csv` (or parquet)
```
annakrystalli, line 163:
```suggestion
   `<round_id>-<model_id>.csv` (or parquet)
```
For consistency with hubverse terminology

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+outfile = os.path.join(path_to_hub, "model-output", "team1-modelA", model_name)
+model_out_tbl.to_csv(outfile, index = False, na_rep = "NA")
+```
+
+(example-parquet)=
+### Example: model output as parquet
+
+Writing to parquet is similar as writing to CSV, but with the caveat that you
+additionally need to ensure that the `output_type_id` column matches the
+[expected `output_type_id_datatype` property of the schema](#output-type-id-datatype).
```
annakrystalli, line 196: That's only if the hub has specified an `output_type_id_datatype`. It might be just auto detected. See https://hubverse.io/en/latest/user-guide/model-output.html#the-output-type-id-column-data-type

This also might need to be performed regardless of the format being written out?. E.g is a column written out from python that contains `NA`s or `None` plus integers still written out as integers?

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+In practice, you will need to know whether or not the expected data type is a
+**string/character** or a **float/numeric**.
```
annakrystalli, line 198: Related to this PR but not expected to be resolved by it, it's clear submitting teams need to have an idea of the hub schema, especally of the `output_type_id` column datatype. While `hubData::create_hub_schema()` and even `hubValidations::submission_tbl()` can provide this info in R, I'm wondering if there's a better way to communicate this infor to submitting teams more broadly?

annakrystalli (changes requested):

docs/source/user-guide/model-output.md
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+
+#### Writing CSV with Python
+
+```python
+import pandas as pd
+import os.path
+# ... generate model data ...
+outfile = os.path.join(path_to_hub, "model-output", "team1-modelA", model_name)
+model_out_tbl.to_csv(outfile, index = False, na_rep = "NA")
+```
```
zkamvar, line 189:
````suggestion
```r
library("fs")
library("readr")
# ... generate model data ...
outfile <- path(hub_path, "model-output", "team1-modelA", model_id)
write_csv(model_out_tbl, outfile)
```

#### Writing CSV with Python

```python
import pandas as pd
import os.path
# ... generate model data ...
outfile = os.path.join(hub_path, "model-output", "team1-modelA", model_id)
model_out_tbl.to_csv(outfile, index = False, na_rep = "NA")
```
````

File: docs/source/user-guide/model-output.md [hunk 17]
```diff
@@ -85,37 +91,135 @@ Per hubverse convention, **there are two groups of columns providing metadata ab
...
+#### Writing parquet with Python
+
+```python
+import pandas as pd
+import os.path
+# ... generate model data ...
+outfile = os.path.join(path_to_hub, "model-output", "team1-modelA", model_name)
+model_out_tbl["output_type_id"] = model_out_tbl["output_type_id"].astype("string") # or "float"
+model_out_tbl.to_parquet(outfile)
+```
```
zkamvar, line 222:
````suggestion

```r
library("fs")
library("arrow")
# ... generate model data ...
outfile <- path(hub_path, "model-output", "team1-modelA", model_id)
model_out_tbl$output_type_id <- as.character(model_out_tbl$output_type_id) # or as.numeric()
arrow::write_parquet(model_out_tbl, outfile)
```


#### Writing parquet with Python

```python
import pandas as pd
import os.path
# ... generate model data ...
outfile = os.path.join(hub_path, "model-output", "team1-modelA", model_id)
model_out_tbl["output_type_id"] = model_out_tbl["output_type_id"].astype("string") # or "float"
model_out_tbl.to_parquet(outfile)
```
````

File: docs/source/user-guide/model-output.md (hunk 14 above)
Comment (zkamvar) [reply docs/source/user-guide/model-output.md:196]: > E.g is a column written out from python that contains NAs or None plus integers still read into R as integers?

Oh good catch! I had updated https://github.com/hubverse-org/hubDocs/issues/198#issuecomment-2430287818 with my findings, but I forgot to update here (in which pandas needs to convert to `Int64` to allow for integer `None` values). 

**The answer to your question is yes**. Even if a parquet file is written with python (either polars or pandas), the type stays with the parquet file and a `null` value inside an integer column is represented as an integer `NA` in R. 

The only thing that seems to have a problem with integer `null` values is pandas. By default, if it reads in a parquet file that has `null` values in an integer column, that column will be coerced to a float (unless the `dtype_backend` is set to `pyarrow`).

File: docs/source/user-guide/model-output.md (hunk 15 above)
Comment (zkamvar) [reply docs/source/user-guide/model-output.md:198]: I think this is exactly the information I needed to know! I was initially confused by the use of `hubData::create_hub_schema()` since I thought it was related to our schemas repository. I had not used these functions and I've learned that some modelers were unaware that hubValidations even existed. At the moment, I think most people are scanning the schema to figure out how to format their data, but the table would likely be helpful and I think it's worthwhile to add into this section. 

As an aside, I wonder if we could add in a function somewhere (maybe in `hubData`) that just provides the expected `output_type_id` datatype given the schema like:

```r
output_type_id_datatype()
#> "numeric"
```

docs/source/user-guide/model-output.md
zkamvar: I'll go ahead and change it. I think when I wrote this, I did not fully realize that `model_out_tbl` was a reserved name.

docs/source/user-guide/model-output.md
zkamvar: I went ahead and iterated on this, rearranging some things and creating examples.

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+Per hubverse convention, **there are two groups of columns providing metadata about the prediction**[^model-id], followed by **a value column with the actual output**. Each group of columns serves a specific purpose: (1) the **"task ID"** columns provide details about what is being predicted, and (2) the two **"model output representation"** columns specify the type of prediction and identifying information about that prediction. Finally, (3) the **value** column provides the model output of the prediction. [Details about the column specifications](column-details) can be found below.
 
 [^model-id]: When using models for downstream analysis with the [`collect_hub()` function](https://hubverse-org.github.io/hubData/reference/collect_hub.html) in the `hubData` package, one more column called `model_id` is prepended added that identifies the model from its filename.
```
annakrystalli, line 91:
```suggestion
[^model-id]: When using models for downstream analysis with the [`collect_hub()` function](https://hubverse-org.github.io/hubData/reference/collect_hub.html) in the `hubData` package, one more column called `model_id` is prepended that identifies the model from its filename. 
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+Per hubverse convention, **there are two groups of columns providing metadata about the prediction**[^model-id], followed by **a value column with the actual output**. Each group of columns serves a specific purpose: (1) the **"task ID"** columns provide details about what is being predicted, and (2) the two **"model output representation"** columns specify the type of prediction and identifying information about that prediction. Finally, (3) the **value** column provides the model output of the prediction. [Details about the column specifications](column-details) can be found below.
 
 [^model-id]: When using models for downstream analysis with the [`collect_hub()` function](https://hubverse-org.github.io/hubData/reference/collect_hub.html) in the `hubData` package, one more column called `model_id` is prepended added that identifies the model from its filename. 
 
+The model output follows the specification of the `tasks.json` configuration
```
annakrystalli, line 93: While the following section is useful and detailed, it feels like it should be more of an aside towards the end of the discussion (i.e. it flows after the section where we tell folks that - btw your missing values need to be typed!). It is also inevitably very R centric when in some ways we're trying to be a little more cross-platform in this section. 

I think it would actually be better to have a way hubs can include schema information in a cross-platform way in their documentation, like a function that prints a table to their README. As I mentioned though this is a separate issue and more generally related to, in my opinion, work we need to do make metadata associated with hub data more visible/accessible.

So for now, I think we should keep this section as it's got lots of useful info but maybe move it down and link it to the mention of typed missing values.

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+arrow schema so that you can ensure your output matches the expected type.** 
+
+Here is some example code that can help. In this example, `tasks_path` is the
+path to your `tasks.json` config file (note: it can also be a URL like [the
+simple forecasts example hub `tasks.json`
+file](https://raw.githubusercontent.com/hubverse-org/example-simple-forecast-hub/refs/heads/main/hub-config/tasks.json))
+
+```r
+# read the configuration file and get the latest round
+config_tasks <- hubData::read_config_file(tasks_path)
```
annakrystalli, line 111: You could use `hubData::read_config(hub_path)` instead and just provide the path to the hub

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+When submitting model output to a hub, it should be placed in a folder with the
+name of your model in the model outputs folder specified by the hub
+administrator (this is usually called `model-output`). Below are two examples
+of writing model output to a hub in R and Python using parquet and CSV files.
+In these examples, we are assuming the following variables already exist:
 
-[^pmf]: **Note on `pmf` model output type**: Values are required to sum to 1 across all `output_type_id` values within each combination of values of task ID variables. This representation should only be used if the outcome variable is truly discrete; a CDF representation is preferred if the categories represent a binned discretization of an underlying continuous variable.
+ - `model_out` is the tabular output from your model formatted as specified
+   in [the formats of model output section](#formats-of-model-output).
+ - `hub_path` is the path to the hub cloned on your local computer
```
annakrystalli, line 229:
```suggestion
 - `hub_path` is the path to the hub cloned on your local computer
 - `model_id` is the combination of `<team_abbr>-<model_abbr>` 
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+To write to CSV, you would use the `write_csv()` from the `readr` package in R and
+the `to_csv()` method in Python. 
+
+#### Writing CSV with R
+
+```r
+library("fs")
+library("readr")
+# ... generate model data ...
+outfile <- path(hub_path, "model-output", "team1-modelA", model_id)
```
annakrystalli, line 245:
```suggestion
outfile <- fs::path(hub_path, "model-output", model_id, file_name)
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+### Example: model output as CSV
 
-[^cdf]: **Note on `cdf` model output type** and `pmf` output type for ordinal variables: In the hub's `tasks.json` configuration file, the values of the `output_type_id` should be listed in order from low to high.
+To write to CSV, you would use the `write_csv()` from the `readr` package in R and
+the `to_csv()` method in Python. 
+
+#### Writing CSV with R
+
+```r
+library("fs")
```
annakrystalli, line 242:
```suggestion
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+library("readr")
```
annakrystalli, line 243:
```suggestion
```
For conciseness

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+# ... generate model data ...
+outfile <- path(hub_path, "model-output", "team1-modelA", model_id)
+write_csv(model_out, outfile)
```
annakrystalli, line 246:
```suggestion
readr::write_csv(model_out, outfile)
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+In practice, you will need to know whether or not the expected data type is a
+**string/character**, **float/numeric**, or an **Int/integer**.
+
+#### Writing parquet with R
+
+```r
+library("fs")
+library("arrow")
+# ... generate model data ...
+outfile <- path(hub_path, "model-output", "team1-modelA", model_id)
```
annakrystalli, line 274:
```suggestion
outfile <- fs::path(hub_path, "model-output", model_id, file_name)
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+Writing to parquet is similar as writing to CSV, but with the caveat that you
+additionally need to ensure that the `output_type_id` column matches the
+[expected `output_type_id_datatype` property of the schema](#output-type-id-datatype).
+In practice, you will need to know whether or not the expected data type is a
+**string/character**, **float/numeric**, or an **Int/integer**.
+
+#### Writing parquet with R
+
+```r
+library("fs")
```
annakrystalli, line 271:
```suggestion
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+library("arrow")
```
annakrystalli, line 272:
```suggestion
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+
+
+(writing-model-output)=
+## Writing model output to a hub
 
-[^batman]: Why have `"NA"` as the `output_type_id`? There are two reasons for this. 
-  First, this provides a placeholder for the model output CSV file in the presence of other output types for validation.
-  The second reason is that we already use `null` to indicate the presence of an absence in the `required` and `optional` fields, and having this allows hubverse tools to treat point estimates without special handling. 
+When submitting model output to a hub, it should be placed in a folder with the
+name of your model in the model outputs folder specified by the hub
```
annakrystalli, line 222:
```suggestion
name of your `model_id` in the model outputs folder specified by the hub
```
`model_id` is not the same as model name. It's the combination of `team_abbr` and `model_abbr`. This is discussed at the top of the page and might be a good idea to link to it.

docs/source/user-guide/model-output.md
annakrystalli: I think we should make it clear that `NA`s might also pop up in task IDs, not just the `output_type_id` column, for example [when a certain task ID is relevant to one modeling task but not another](https://hubverse.io/en/latest/quickstart-hub-admin/tasks-config.html#required-and-optional-elements) (also discussed in https://github.com/hubverse-org/hubValidations/issues/140)

annakrystalli (changes requested):
Getting there from my perspective with some really useful and detailed additions!

Made a few corrections and a suggestion for moving a section and cleaning up some possible misconceptions.

docs/source/user-guide/model-output.md
zkamvar: I was thinking about @trobacker's comment that he usually models outside of the context of a hub, but I can certainly change this.

docs/source/user-guide/model-output.md
zkamvar: Oh good point! I did not realize this!

docs/source/user-guide/model-output.md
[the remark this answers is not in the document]
zkamvar: 🙈

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,204 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+
+The model output type IDs have different caveats depending on the `output_type`:
+
+`mean` and `median`
+: Point estimates do not have an `output_type_id` because you can only have one
+point estimate for each combination of task IDs. However, because the
+`output_type_id` column is required, something has to go in this place, which
+is a missing value. This is encoded as [`NA` in
+R](https://www.njtierney.com/post/2020/09/17/missing-flavour/) (which is why
+our schemas prior to 4.0.0 encoded these as `["NA"]`) and `None` in Python. See
```
bsweger, line 192:
```suggestion
R](https://www.njtierney.com/post/2020/09/17/missing-flavour/) and `None` in Python. See
```
I think this reads better without the aside about schema versions...the important thing is to convey what the `output_type_id` column will look like for point estimate tasks

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+```
 
+#### Writing CSV with Python
+
+```python
+import pandas as pd
+import os.path
+# ... generate model data ...
+outfile = os.path.join(hub_path, "model-output", "team1-modelA", model_id)
+model_out.to_csv(outfile, index = False, na_rep = "NA")
```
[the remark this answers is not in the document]
zkamvar: Yeah it's nuanced and annoying. I should probably just leave it out and allow NA's to be blanks. R should be able to handle them.

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,204 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+```
+
+(example-parquet)=
+### Example: model output as parquet
+
+Writing to parquet is similar as writing to CSV, but with the caveat that you
+additionally need to ensure that the `output_type_id` column matches the
+[expected `output_type_id_datatype` property of the schema](#output-type-id-datatype).
```
bsweger, line 264:
```suggestion
Unlike a CSV, a parquet files contain embedded information about the data types of its
columns. Therefore, when writing model output files as parquet, it's critical that you
first ensure the data type of the `output_type_id` column matches the
[expected `output_type_id_datatype` property of the schema](#output-type-id-datatype).

If the data types of the model output parquet file don't match the hub's schema, the
submission will not validate.
```

Tried to surface the reason for the difference between csv and parquet, without getting too in-the-weeds. Also think it might be helpful to clarify why it's important to mind the parquet column types (maybe that would be a good message for one of those fancy warning boxes?)

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,204 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+
+
+#### Writing parquet with Python
+
+```python
+import pandas as pd
+import os.path
+# ... generate model data ...
+outfile = os.path.join(hub_path, "model-output", "team1-modelA", model_id)
+model_out["output_type_id"] = model_out["output_type_id"].astype("string") # or "float", or "Int64"
```
bsweger, line 285:
```suggestion
# update the output_type_id data type to match the hub's schema
model_out["output_type_id"] = model_out["output_type_id"].astype("string") # or "float", or "Int64"
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,204 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+[expected `output_type_id_datatype` property of the schema](#output-type-id-datatype).
+In practice, you will need to know whether or not the expected data type is a
+**string/character**, **float/numeric**, or an **Int/integer**.
+
+#### Writing parquet with R
+
+```r
+# ... generate model data ...
+outfile <- fs::path(hub_path, "model-output", model_id, file_name)
+model_out$output_type_id <- as.character(model_out$output_type_id) # or as.numeric(), or as.integer()
```
bsweger, line 273:
```suggestion
# update the output_type_id data type to match the hub's schema
model_out$output_type_id <- as.character(model_out$output_type_id) # or as.numeric(), or as.integer()
```

bsweger (commented):
Thanks for taking the time to make sure I'm understanding the problem space correctly! Based on that, I made a few more wording suggestions for your consideration (feel free to disregard)

docs/source/user-guide/model-output.md
```diff
@@ -46,7 +46,11 @@ one-week-ahead incidence, but probabilities for the timing of a season peak:
 | EW202242 | weekly rate | 1 | sample | 2 | 3 |
 :::
 
-(formats-of-model-output)=
+[^batman]: `NA` (without quotes) indicates missingness in R, which is the expected `output_type_id` for a `mean` `output_type`.
```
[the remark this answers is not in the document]
zkamvar: This might lead people to encode `None` in the CSV itself, which would be interpreted as a character. What about this: 

```suggestion
[^batman]: The `output_type_id` for point estimates (e.g. `mean`) is not applicable. To reflect this, we need to signal that this is a missing value. While a blank cell could indicate missingness, we chose `NA` because it's a clear indicator that it's a presence of an absence and not a mistake. In R, missing values are encoded as `NA` and in Python, they are encoded as `None`.
```

docs/source/user-guide/model-output.md
zkamvar: totes agree

docs/source/user-guide/model-output.md
zkamvar: I think this is much better wording!

LucieContamin (commented):
Thank you for the update, it looks all clear to me, I just made some minor comments.

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,204 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+
+The model output follows the specification of the `tasks.json` configuration
+file of the hub. If you are creating a model and you would like to know what
+data type your columns should be in, the hubVerse has utilities to provide and
+arrow schema and even a full submission template from the `tasks.json`
+configuration file. 
+
+### Arrow Schema
+
+**The hubverse package `hubData()` has functionality that will generate an
```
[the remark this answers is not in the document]
zkamvar: good catch!

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +86,207 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+simple forecasts example hub `tasks.json`
+file](https://raw.githubusercontent.com/hubverse-org/example-simple-forecast-hub/refs/heads/main/hub-config/tasks.json))
+
+```r
+# read the configuration file and get the latest round
+config_tasks <- hubData::read_config_file(tasks_path)
+rounds <- hubData::get_round_ids(config_tasks)
+this_round <- rounds[length(rounds)]
+# create the submission template (this may take some time if your submission uses samples)
+tmpl <- hubValidations::submission_tmpl(config_tasks, round_id = this_round)
```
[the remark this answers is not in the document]
zkamvar: Thank you! The assertion on `hub_con` was actually a failure from my earlier code where I did not name the argument.

docs/source/user-guide/model-output.md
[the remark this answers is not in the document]
zkamvar: I've looked at this again and I've removed the `na_rep` argument to simplify the example and it works.

docs/source/user-guide/model-output.md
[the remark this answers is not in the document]
bsweger: > This might lead people to encode `None` in the CSV itself, which would be interpreted as a character. What about this:

I take your point! But I find this confusing (perhaps because it's R-centric). Maybe less is more?
```suggestion
[^batman]: The `output_type_id` for point estimates (e.g. `mean`) is not applicable. To
reflect this, we need to signal that this is a missing value. In R, missing values are
encoded as `NA`, and in Python, they are encoded as `None`.
```

docs/source/user-guide/model-output.md
zkamvar: Yeah, I wasn't super happy with the placement of this and I agree with your sentiment. This is further evidence I need to split the documentation into the Diátaxis framework. 

> I think it would actually be better to have a way hubs can include schema information in a cross-platform way in their documentation, like a function that prints a table to their README. As I mentioned though this is a separate issue and more generally related to, in my opinion, work we need to do make metadata associated with hub data more visible/accessible.

Y'know, I wouldn't be surprised if this could be achieved with some vanilla in-browser JS.

docs/source/user-guide/model-output.md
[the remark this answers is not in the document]
zkamvar: Sold!

zkamvar:
Thank you @annakrystalli, @bsweger, @LucieContamin, and @micokoch for your comments!

As per @annakrystalli and @LucieContamin's suggestions, I have moved the documentation about the template generation and arrow schema closer to the examples. Specifically, I added the section about the arrow schema _inside_ the example so that users do not have to look far to know what the arrow schema would look like.

As per @bsweger's suggestions (and https://github.com/hubverse-org/hubDocs/issues/198#issuecomment-2448127804), I have attempted to focus the discussion for the modelers on the column type.

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +88,229 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+If the data types of the model output parquet file don't match the hub's schema, the
+submission will not validate.
+In practice, you will need to know whether or not the expected data type is a
+**string/character**, **float/numeric**, or an **Int/integer**.
+
+(arrow-schema)=
+#### Arrow Schema
 
-[^pmf]: **Note on `pmf` model output type**: Values are required to sum to 1 across all `output_type_id` values within each combination of values of task ID variables. This representation should only be used if the outcome variable is truly discrete; a CDF representation is preferred if the categories represent a binned discretization of an underlying continuous variable.
+**The hubverse packages `hubData()` and `hubUtils()` have functionality that will generate an
```
annakrystalli, line 263:
```suggestion
**The hubverse packages `hubData` and `hubUtils` have functionality that will generate an
```

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +88,229 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+ - `model_id` is the combination of `<team_abbr>-<model_abbr>`
+ - `file_name` is the file name of your model formatted as
+   `<round_id>-<model_id>.csv` (or `.parquet`)
+ - `model_out` is the tabular output from your model formatted as specified
+   in [the formats of model output section](#formats-of-model-output).
+
+(submission-template)=
+### Submission Template
+
+**The hubverse package `hubValidations()` has functionality
```
annakrystalli, line 184:
```suggestion
**The hubverse package `hubValidations` has functionality
```
As its a pkg name and not a function name, we don't need parentheses

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +88,229 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+target: string
+horizon: int32
+location: string
+age_group: string
+output_type: string
+output_type_id: double
+value: int32
+model_id: string
+```
+
```
annakrystalli, line 289:
```suggestion

If you have a data frame of model output data (e.g. `model_out`) and want to coerce the column data types to the hub schema, you can also use [`hubData::coerce_to_hub_schema()`](https://hubverse-org.github.io/hubData/reference/coerce_to_hub_schema.html), e.g. 
```r
hubData::coerce_to_hub_schema(model_out, config_tasks)
```

```

**Review by annakrystalli — APPROVED** [review]
I made a couple of minor last minute corrections and added a suggested note about another useful function, `hubData::coerce_to_hub_schema()`.

Other than that, all my comments have been resolved. Thanks @zkamvar !

**Comment by annakrystalli** [comment]
FYI as a response to functions being relevant to modelers appearing across too many packages, I've made a proposal to re-export them to `hubValidations` here: https://github.com/hubverse-org/hubValidations/issues/149

Input welcome

File: docs/source/user-guide/model-output.md (hunk 19 above)
Comment (annakrystalli) [reply docs/source/user-guide/model-output.md]: Oooh that could be cool! Maybe we should open an issue somewhere to track this? (Not sure where the best place would be...)

File: docs/source/user-guide/model-output.md (hunk 20 above)
Comment (annakrystalli) [reply docs/source/user-guide/model-output.md]: They still need access to the hub to get to the config file though? The only difference between the two functions is that one takes the full path to the config file and the other the path to a hub and then builds the config path relative to the root of the hub. It also defaults to the `tasks.json` which is the config most commonly loaded.

File: docs/source/user-guide/model-output.md (hunk 39 above)
Comment (LucieContamin) [reply docs/source/user-guide/model-output.md:184]: I nearly suggested it but was not sure I was correct, only function need parentheses, no? (Just to be sure I have the correct standard)

File: docs/source/user-guide/model-output.md [hunk 41]
```diff
@@ -78,44 +88,229 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+#### Writing parquet with R
+
+```r
+# ... generate model data ...
+outfile <- fs::path(hub_path, "model-output", model_id, file_name)
 
+# update the output_type_id data type to match the hub's schema
+model_out$output_type_id <- as.character(model_out$output_type_id) # or as.numeric(), or as.integer()
```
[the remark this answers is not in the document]
zkamvar: oh good point!

docs/source/user-guide/model-output.md
```diff
@@ -78,44 +88,229 @@ Much of the material in this section has been excerpted or adapted from the [hub
...
+#### Writing parquet with Python
+
+```python
+import pandas as pd
+import os.path
+# ... generate model data ...
+outfile = os.path.join(hub_path, "model-output", model_id, file_name)
+
+# update the output_type_id data type to match the hub's schema
+model_out["output_type_id"] = model_out["output_type_id"].astype("string") # or "float", or "Int64"
```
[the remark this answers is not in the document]
zkamvar: thank you!

docs/source/user-guide/model-output.md
zkamvar: I opened it in the site builder dashboard repository.

zkamvar:
Thank you all for sticking with me through this!

Issue #200, opened 2024-10-22
Catalog abbreviations used in packages and documentation


zkamvar:
Something that occasionally trips me up is the use of abbreviations in code and I realized I needed a way to keep track of what abbreviations were used and how they were used, so I wrote up a table to track them and I think this would be a good appendix to have. That being said, I'm not sure if I got everything.

| abbreviation | meaning 
| ------------ | ------- |
| `abbr`       | abbreviation |
| `cdf`        | cumulative density function |
| `cdn`        | condition |
| `cfg`        | config/configuration |
| `col`        | column |
| `eq`         | equal |
| `exec`       | execution |
| `fn`         | function |
| `gt`         | greater than |
| `gte`        | greater than or equal |
| `idx`        | index (of an array) |
| `lt`         | less than |
| `lte`        | less than or equal |
| `mt`         | model task |
| `out`        | output |
| `pkg`        | (R) package | 
| `pmf`        | probability mass function |
| `popn`       | population |
| `pr`         | pull request |
| `prop`       | property/proportion (depending on context) |
| `spl`        | sample (from a distribution) |
| `taskid`     | task ID |
| `tbl`        | table/tibble/data frame |
| `tid`        | output type ID |
| `tmpl`       | template |
| `ts`         | time series |
| `val`        | value |

PR #206, opened 2024-11-06
Add abbreviations page

micokoch:
resolves issue #200

### DIFF
--- a/docs/source/overview/abbreviations.md
+++ b/docs/source/overview/abbreviations.md
@@ -1,0 +1,34 @@
+# Abbreviations
+
+The hubverse code often has abbreviations which may not be familiar to users. This page provides a list of abbreviations and their meanings. Let us know if we missed any by filing an issue on the [hubDocs GitHub repository](https://github.com/hubverse-org/hubDocs/issues).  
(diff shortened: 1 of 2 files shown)


docs/source/overview/abbreviations.md
```diff
@@ -0,0 +1,34 @@
+# Abbreviations
+
+The hubverse code often has abbreviations which may not be familiar to users. This page provides a list of abbreviations and their meanings. Let us know if we missed any by filing an issue on the [hubDocs GitHub repository](https://github.com/hubverse-org/hubDocs/issues).
```
zkamvar, line 3:
```suggestion
The hubverse code often has abbreviations which may not be familiar to users. This page provides a list of abbreviations and their meanings. Let us know if we missed any by [filing an issue on the hubDocs GitHub repository](https://github.com/hubverse-org/hubDocs/issues).  
```

zkamvar (approved):
Thank you! I have one suggested change. Additionally, in the future, these kind of reference pages should not live in the overview, but rather an appendix since no one is going to read this table from top to bottom.

PR #208, opened 2024-11-06
Update Target Data for updated data standards

### LINKED ISSUE #204: migrate Target Data formats writeup to hubDocs
@elray1 wrote up [documentation for target data formats](https://htmlpreview.github.io/?https://github.com/reichlab/hub-infrastructure-experiments/blob/main/eval-examples/target_data_formats.html). The [source file](https://github.com/reichlab/hub-infrastructure-experiments/blob/main/eval-examples/target_data_formats.Rmd) is an R Markdown document.

I will move this to the documentation and translate the tibble output to tables.

zkamvar:
I modified @elray1's [writeup of proposed target data standards](https://htmlpreview.github.io/?https://github.com/reichlab/hub-infrastructure-experiments/blob/main/eval-examples/target_data_formats.html) ([source](https://github.com/reichlab/hub-infrastructure-experiments/blob/7cf7f3f9fcd9e2cd492a51ab1de0d823ec47fadc/eval-examples/target_data_formats.Rmd)) a bit to show apples to apples comparisons while respecting the limited space and had it output to GitHub document which I then transferred over here. 

I also made the executive decision to overwrite the original targets page. 

**Preview of changes: https://hubdocs--208.org.readthedocs.build/en/208/user-guide/target-data.html**

Things I changed from Evan's version:

1. The data are presented in a consistent manner, showing outputs from Massachusettes across four time points. 
2. The data are presented as tables, not tibbles
3. The tables have columns trimmed to fit the page width
4. I added a figure that demonstrates the relationship between model distribution and oracle distribution (this can be modified)
1. I revamped the oracle output details section from one long list to two short subsections that align with the model outputs chapter terminology. I did this because there were really long explanations to describe that oracle outputs would contain only independent task ID variables and that the output type representation columns were only necessary for PMF and CDF output types. Moreover, I removed the language describing example scenarios from that section as the examples do a better job at demonstrating the
scenarios. 

This addresses the conversation that has occurred in https://github.com/orgs/hubverse-org/discussions/9 and will fix #204


docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,510 @@
...
+
+A hub will typically have access to data in time series format, and will
+need to convert it to the **oracle output** format for use with any tools
+that require it in that format (see the next section). In hubs that
+collect mean, median, quantile, or sample predictions for the reported
+signal values in the raw time series data, the two formats may be
+essentially the same, perhaps with some renaming of columns. However, in
+hubs that form predictions for quantities that are derived from the the
+raw time series data, such as the peak time or peak incidence, and in
+hubs that collect pmf or cdf predictions, the formats will differ more
```
LucieContamin, line 146: Should we add one or two sentences to indicate more information below if available and link it to you? 
Or reorganize the file to have the "more about" directly after?

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,510 @@
...
+    and/or oracle output programmatically
+2.  by storing snapshots of the target time series data and/or oracle
+    output in the hub repository
+
+Following general conventions for storage of code related to modeling
+hubs, we recommend that any code for data access be provided in a
+separate repository following standard language-specific packaging
+guidelines, or if the code is small in scope it can be placed within the
+`src` folder of the hub’s repository.
+
```
LucieContamin, line 189: Should we add recommendation on where to store the data? For example in a `target-data` folder?

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,510 @@
...
+      `output_type_id` corresponds to the observed category (indicating a
+      probability of 1 for that category) and `0` for other categories.
+    - For the `cdf` output type, the `oracle_value` is `0` for any
+      `output_type_id` levels that are less than the observed value, and `1`
+      for any `output_type_id` levels that are greater than or equal to the
+      observed value, corresponding to the step function cdf of a
+      probability distribution that places all of its probability at the
+      observed value.
+
+### Oracle output vs model output
```
LucieContamin, line 221: Should we put the information a little bit higher, I think it helps a lot with understanding why we design it that way. For example, at the beginning of the "Oracle output" section.

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,510 @@
...
+The model output can contain any number of task ID columns that are used to
+provide details about what is being predicted based on
+
+ - independent task ID variables (e.g. `location`, `target_date`, and
+   `age_group`),
+ - derived (dependent on other variables) task ID variables (e.g. `horizon`,
+   `reference_date`, `origin_date`), and
+ - scenario-specific task ID variables (e.g. `scenario_id`).
+
+**The oracle output will contain the independent task ID variables** that are
```
LucieContamin, line 252: Do you mean "independent and any other variables" necessary to match?
Because you will need some date/time information probably and with this list, it seems that all time related variable are called "derived".

LucieContamin (commented):
Looking good! thank you very much. I just made some comments mainly minor suggestions to re-organize or add more information.

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,510 @@
...
+- Predictions correspond to a distribution that places probability 1 on
+  the observed target outcome (see figure below).
+- Predictions (e.g., means, quantile values, or pmf category
+  probabilities, etc.) are stored in a column named `oracle_value`
+  rather than `value`.
+- We introduce some conventions to avoid duplication of data, described
+  in more detail below. Generally, the columns of the oracle output will
+  be a subset of the columns of valid model output for the hub, with
+  just those columns that are needed to correctly align `oracle_value`s
+  with the corresponding predicted `value`s produced by modelers.
```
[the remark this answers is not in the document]
zkamvar: Good catch! Thank you!

docs/source/user-guide/target-data.md
zkamvar: Yes! Thank you for mentioning this! In fact, we should have recommendations for standardized file names (unless we want to have a way to specify this in the admin schema).

docs/source/user-guide/target-data.md
zkamvar: I think that's a good point. I'll move it up.

docs/source/user-guide/target-data.md
zkamvar: Probably. I was trying to avoid over-complicating the statement. Thinking about it, the oracle data should have the same columns as the time series data (except `date` is renamed to `target_date`) with additional columns that describes the metadata about the outputs.... unless I'm completely off-base? 

I was attempting to summarize @elray1's original text:

>  - The oracle output should include enough of the task id variables and columns with metadata about the outputs (output_type and output_type_id) to uniquely identify which oracle_values correspond to which predicted values. For example, this will typically include task id variables such as location and target_date (or target_end_date), since the oracle_value will be specific to the location and target_date.
>  - Any task id variables that are not necessary to match observations with predictions can be omitted from the oracle output. For example, if target_date is included then reference_date and horizon variables can be omitted because the same observation will generally correspond to a particular target_date regardless of the forecast horizon. Similarly, in a scenario projection setting, the scenario_id can be omitted.

docs/source/user-guide/target-data.md
zkamvar: Agreed this is a bit hedging, so I've moved it before the description of the column types so that it flows better.

zkamvar:
In thinking about this, I realize that there are still gaps in defining the standard for time series data. We mention that it should be in long format (albeit with many more words) and that it should have an "observation" column and a time index column. 

From what I can see, there still needs to be a bit more structure to having something that we can use to consistently read in the target data. I can see a few ways of addressing this:

1. mandate specific column and file names
2. add configurations to the admin schema that defines the path to the time series and oracle output data, mapping column names to targets.
3. the same as 2, but having a specific `targets.json` spec.

docs/source/user-guide/target-data.md
LucieContamin: For me the original text is clear enough. I will be ok with keeping it.

zkamvar: The original text feels a lot like legal jargon and I had a hard time understanding what it actually means.

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,519 @@
...
+  or `cdf` output types, the `output_type` and `output_type_id` columns
+  could be omitted and this duplication could be eliminated.
+
+:::{note}
+
+These examples are all collected and filtered from [the `hubExamples` package](https://hubverse.github.io/hubExamples). The model output data set contains over
+10,000 rows and the oracle output data has over 200,000 rows.
+
+To make comparisons easier, we have subset the data to Massachusettes (FIPS code
+25) with four target end dates between 2022-11-19 and 2022-12-10.
```
elray1, line 294:
```suggestion
To make comparisons easier, we have subset the data to Massachusettes (FIPS code
25) with one `reference_date` of 2022-11-19 and four target end dates between 2022-11-19 and 2022-12-10.
```

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,519 @@
...
+which is the quantile of a predictive distribution that assigns
+probablity 1 to that observed value at any quantile probability level. A
+model output file would need to have a separate row for each quantile
+level reported in the `output_type_id` column. As a space-saving
+convention, we use `output_type_id = <NA>` to indicate that this
+`oracle_value` applies to all quantile levels.
+
+### Output type `sample`
+
+:::{table} A subset of **model output** showing 6 `sample` predictions across one horizon
```
elray1, line 397:
```suggestion
:::{table} A subset of **model output** showing 6 `sample` predictions at one horizon
```

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,519 @@
...
+| 25 | 2022-12-03 | wk flu hosp rate category | pmf | low | 0 |
+| 25 | 2022-12-03 | wk flu hosp rate category | pmf | moderate | 0 |
+| 25 | 2022-12-03 | wk flu hosp rate category | pmf | high | 1 |
+| 25 | 2022-12-03 | wk flu hosp rate category | pmf | very high | 0 |
+:::
+
+The presence of a `1` for the `oracle_value` in the first row and `0` in the
+subsequent three rows indicates that the observed rate category in
+Massachusettes on the week of 2022-11-19 was `"low"` while the week of
+2022-11-26 was `"moderate"`.
```
elray1, line 464:
```suggestion
The presence of a `1` for the `oracle_value` in the first row and `0` in the
subsequent three rows indicates that the observed rate category in
Massachusettes on the week of 2022-11-19 was `"low"`. Similarly, the observed rate category for the week of
2022-11-26 was `"moderate"`.
```

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,519 @@
...
+scenario projection setting, the `scenario_id` can be omitted as there is only
+one scenario for an observed event[^quanta].
+
+[^quanta]: just don't tell the quantum physicists.
+
+### Model output representation columns
+
+**The `output_type` and `output_type_id` columns only need to be included if
+the hub collects `pmf` or `cdf` outputs.** As was described above, for those
+two output types the `oracle_value` depends on the `output_type_id`. On the
```
elray1, line 208:
```suggestion
**The `output_type` and `output_type_id` columns only need to be included if
the hub collects `pmf` or `cdf` outputs.** For those two output types the
`oracle_value` depends on the `output_type_id` (see the next section for more
detail). On the
```

docs/source/user-guide/target-data.md
```diff
@@ -1,64 +1,519 @@
...
+A hub will typically have access to data in time series format, and will
+need to convert it to the **oracle output** format for use with any tools
+that require it in that format (see the next section). In hubs that
+collect mean, median, quantile, or sample predictions for the reported
+signal values in the raw time series data, the two formats may be
+essentially the same, perhaps with some renaming of columns. However, in
+hubs that form predictions for quantities that are derived from the the
+raw time series data, such as the peak time or peak incidence, and in
+hubs that collect pmf or cdf predictions, the formats require the presence of
+model output representation columns.
```
elray1, line 180:
```suggestion
A hub will typically have access to data in time series format, and will
need to convert it to the **oracle output** format for use with any tools
that require it in that format (see the next section). In hubs that
collect mean, median, quantile, or sample predictions for the reported
signal values in the raw time series data, the two formats may be
essentially the same, perhaps with some renaming of columns. However,
these data formats will differ more in hubs that form predictions for quantities
that are derived from the the raw time series data, such as the peak time or
peak incidence, and in hubs that collect pmf or cdf predictions.
```

elray1: See a separate comment on the PR for more about my thinking here.  I decided that maybe simpler was better in terms of just stating that the formats would differ in these situations rather than providing more detailed examples.  I'm open to further revision though!

elray1 (changes requested):
Looks great!  I'm requesting a few minor changes.

docs/source/user-guide/target-data.md
zkamvar: I think the "for" was a typo in there and can be removed unless we want to have it sound more prophetic than it should.

docs/source/user-guide/target-data.md
zkamvar: I think this is definitely moving in the right direction!

elray1 (approved):
looks good! thanks!

PR #215, opened 2024-11-14
collapse navbar elements

zkamvar:
This [collapses the navbar elements](https://sphinx-book-theme.readthedocs.io/en/stable/sections/sidebar-primary.html#control-the-depth-of-the-left-sidebar-lists-to-expand) according to the book theme docs.

It does what it says on the tin: it collapses the navbar elements, but the downside is now in order to get anywhere, you need to click twice to open the navbar and then go to the link you wanted to go to.


zkamvar:
Thank you, @micokoch!

PR #216, opened 2024-11-21
Adding a dynamic contributors page to hubverse site

micokoch:


micokoch:
Hi @annakrystalli and @zkamvar. As I mentioned on Slack, this is a very rough draft. Let me know if this works for you (and I can improve on it) or if you'd rather I kill this idea (maybe because you don't want more scripts, or it interferes with the workflow, or something else). I won't work on it until next week, so there's no hurry.

micokoch:
I'm going to leave this here for consideration. The Python script would have to be added to a workflow, and the layout itself could use feedback, but there's not much more I want to improve on it for now.
[Preview](https://hubdocs--216.org.readthedocs.build/en/216/overview/contributors.html).

scripts/update_contributors.py
```diff
@@ -0,0 +1,79 @@
...
+                else:
+                    blog_output = ""  # Don't include parentheses if `blog` is empty
+
+                # Avoid adding period if `bio` or `location` is empty
+                bio_output = f"{bio}." if bio else ""
+                location_output = f" {location}." if location else ""
+
+                # Write to the file with conditional formatting
+                file.write(
+                    f'<img src="{avatar_url}" alt="{name or contributor["login"]}" class="avatar"> '
```
zkamvar, line 67: since these images are more decorative than informative, it's best to leave alt blank so that screen readers knows to skip over them.

```suggestion
                    f'<img src="{avatar_url}" alt="" class="avatar"> '
```

scripts/update_contributors.py
```diff
@@ -0,0 +1,79 @@
...
+                    f"- {name_output}{blog_output} ([{contributor['login']}]({contributor['html_url']})) - "
+                    f"{bio_output}{location_output} {commit_count} commits.\n\n---\n\n"
+                )
+            else:
+                file.write(
+                    f"- ![Avatar](https://dummyimage.com/50x50/3c88be/3c88be) "
```
zkamvar, line 73: IMO, https://placekittens.com >>> https://dummyimage.com

```suggestion
                    f"- ![Avatar](https://placekittens.com/g/50/50) "
```

zkamvar (approved):
I've made a couple of suggestions to improve accessibility.

scripts/update_contributors.py
micokoch: Thanks for the suggestions, but when I try to go to the placekittens link you post, it doesn't work.

zkamvar: > Thanks for the suggestions, but when I try to go to the placekittens link you post, it doesn't work.

This doesn't work? Do you get a 404? https://placekittens.com/g/50/50

PR #217, opened 2024-11-27
add derived task ID language

### LINKED ISSUE #193: Add documentation about `derived_task_ids` property
When updating to  `v4.0.0` add short section about derived taks IDs

Could be a rephrasing of the NEWS entry in schema:

> Introduction of optional `derived_task_ids` properties to enable hub administrators to define derived task IDs (i.e. task IDs whose values depend on the values of other task IDs). The higher level `derived_task_ids` property sets the property globally at the hub level but can be overriden by the round level `derived_task_ids` property. The property allows for primarily validation functionality to ignore such task IDs when appropriate which can significantly improve validation efficency (#96). For more information see [`hubValidations` documentation on ignoring derived task IDs](https://hubverse-org.github.io/hubValidations/articles/validate-pr.html#ignoring-derived-task-ids-to-improve-performance).

zkamvar:
This will fix #193 by adding language to describe derived task IDs.

### DIFF
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -99,7 +99,7 @@ check failure. The table below shows an expanded grid of `horizon`, `origin_date
 ```{table} validation will provide false errors if there are unspecified derived task IDs
 
 | `origin_date` | `horizon` | `target_date` | valid | `derived_task_ids:` `null` | `derived_task_ids:` `["target_date"]` |
-| ----------    | --------- | ------------- | ----------------- | ------------------------ | ----------------------------------- |
+| ------------- | --------- | ------------- | ----- | ------------------------ | ----------------------------------- |
 | 2024-11-07 | 1 | 2024-11-14 | 🆗 | ✅ | ✅ |
 | 2024-11-07 | 2 | 2024-11-21 | 🆗 | ✅ | ✅ |
 | 2024-11-07 | 1 | 2024-11-21 | 🚫 | ❌ | NA |
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -82,24 +82,41 @@ significantly improve validation efficency. For more information see the
 IDs](https://hubverse-org.github.io/hubValidations/articles/validate-pr.html#ignoring-derived-task-ids-to-improve-performance).
 
 :::{note}
-If any task IDs with `required` values have an associated derived task ID, **it is essential for `derived_task_ids` to be specified**. Otherwise, this will result in false validation errors. For example, if you have two required `horizon` values of 1 and 2, then you _must_ include `target_date` in `derived_task_ids`.
-Without specifying this, you will end up with a
+If any `required` task IDs have an associated derived task ID, **it is essential for `derived_task_ids` to be specified**. Otherwise, this will result in false validation errors. 
+
+Take for example a scenario where  `target_date` is derived from `origin_date`
+and `horizon` via `target_date = origin_date + horizon * 7`. If you have a required `origin_date` value of "2024-11-07", then
+you _must_ include `target_date` in `derived_task_ids`. Without specifying
+this, you will end up with a
 [`req_vals`](https://hubverse-org.github.io/hubValidations/reference/check_tbl_values_required.html)
-check failure. The table below shows an expanded grid of `horizon`, `origin_date`, and `target_date` for a single `origin_date` and two `horizons` with the results of the `req_vals` check if `derived_task_ids` is unspecified (null) or set to `target_date`:
+check failure. 
 
-```{table} validation will provide false errors if there are unspecified derived task IDs
+The tables below show the validation grid of `horizon`, `origin_date`, and `target_date` for a single required `origin_date` and two `horizons` in the scenarios where `derived_task_ids` is unspecified (null) or set to `"target_date"`.
 
-| `origin_date` | `horizon` | `target_date` | valid | `derived_task_ids:` `null` | `derived_task_ids:` `["target_date"]` |
-| ------------- | --------- | ------------- | ----- | -------------------------- | ------------------------------------- |
-| 2024-11-07 | 1 | 2024-11-14 | 🆗 | ✅ | ✅ |
-| 2024-11-07 | 2 | 2024-11-21 | 🆗 | ✅ | ✅ |
-| 2024-11-07 | 1 | 2024-11-21 | 🚫 | ❌ | NA |
-| 2024-11-07 | 2 | 2024-11-14 | 🚫 | ❌ | NA |
+This validation grid causes a validation failure because 2024-11-21 is NOT 1 week ahead of 2024-11-07 and 2024-11-14 is NOT 2 weeks ahead of 2024-11-07.
 
+```{table} With <code>"derived_task_ids": null</code>, validation will provide false errors.
+:width: 95%
+
+| `origin_date` | `horizon` | `target_date` | validation result |
+| ------------- | --------- | ------------- | ----------------- |
+| 2024-11-07 | 1 | 2024-11-14 | ✅ |
+| 2024-11-07 | 2 | 2024-11-21 | ✅ |
+| 2024-11-07 | 1 | 2024-11-21 | ❌ |
+| 2024-11-07 | 2 | 2024-11-14 | ❌ |
 ```
 
-As shown in the last column of the above table, the `req_vals` validation passes when `derived_task_ids` is set to `["target_date"]`.
+This expanded grid will validate correctly because `"target_date"` has been identified as a derived task ID and is appropriately masked before the validation grid is greated. 
 
+```{table} By setting <code>"derived_task_ids": ["target_date"]</code>, validation succeeds
+:width: 95%
+
+| `origin_date` | `horizon` | `target_date` | validation result |
+| ------------- | --------- | ------------- | ----------------- |
+| 2024-11-07 | 1 | `<NA>` | ✅ |
+| 2024-11-07 | 2 | `<NA>` | ✅ |
+
+```
 
 :::
 
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -89,34 +89,41 @@ and `horizon` via `target_date = origin_date + horizon * 7`. If you have a requi
 you _must_ include `target_date` in `derived_task_ids`. Without specifying
 this, you will end up with a
 [`req_vals`](https://hubverse-org.github.io/hubValidations/reference/check_tbl_values_required.html)
-check failure. 
+check failure _even if your submission file is valid_.
 
-The tables below show the validation grid of `horizon`, `origin_date`, and `target_date` for a single required `origin_date` and two `horizons` in the scenarios where `derived_task_ids` is unspecified (null) or set to `"target_date"`.
+If your submission file looks like this: 
 
-This validation grid causes a validation failure because 2024-11-21 is NOT 1 week ahead of 2024-11-07 and 2024-11-14 is NOT 2 weeks ahead of 2024-11-07.
-
-```{table} With <code>"derived_task_ids": null</code>, validation will provide false errors.
+```{table} submission file
 :width: 95%
+:widths: grid
+| `origin_date` | `horizon` | `target_date` |        ...        |
+| ------------- | --------- | ------------- | ----------------- |
+| 2024-11-07 | 1 | 2024-11-14 | ... |
+| 2024-11-07 | 2 | 2024-11-21 | ... |
+```
 
+**When you set `"derived_task_ids": ["target_date"]`, then your submission will
+pass the validation checks ✅.**
+
+However, **without setting `derived_task_ids`, you will get an ❌
+`<error/check_failure>`** with a message indicating that required task ID
+combinations are missing. The "missing" combintaions are shown in the table
+below.
+
+```{table} **<code><error/check_failure></code>** With <code>"derived_task_ids": null</code>, validation will provide false errors.
+:width: 95%
+:widths: 2, 2, 1, 2
 | `origin_date` | `horizon` | `target_date` | validation result |
 | ------------- | --------- | ------------- | ----------------- |
-| 2024-11-07 | 1 | 2024-11-14 | ✅ |
-| 2024-11-07 | 2 | 2024-11-21 | ✅ |
 | 2024-11-07 | 1 | 2024-11-21 | ❌ |
 | 2024-11-07 | 2 | 2024-11-14 | ❌ |
 ```
 
-This expanded grid will validate correctly because `"target_date"` has been identified as a derived task ID and is appropriately masked before the validation grid is greated. 
+If you inspect the table, you will notice that _these combinations are invalid_
+because the values in `target_date` are not correctly aligned with the 
+`origin_date` and `horizon` (i.e. 2024-11-21 is _two weeks_ ahead of
+2024-11-07, not 1 week, as indicated by the `horizon`).
 
-```{table} By setting <code>"derived_task_ids": ["target_date"]</code>, validation succeeds
-:width: 95%
-
-| `origin_date` | `horizon` | `target_date` | validation result |
-| ------------- | --------- | ------------- | ----------------- |
-| 2024-11-07 | 1 | `<NA>` | ✅ |
-| 2024-11-07 | 2 | `<NA>` | ✅ |
-
-```
 
 :::
 
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -64,17 +64,21 @@ In this example, `target` is the target key and can only take on one value, "inc
 #### Derived Task ID variables
 
 Each model output is defined on the unique combinations across the task ID
-variables. For example, across 10 `origin_date`s and 50 `location`s, there are
-500 unique combinations of tasks because both of these variables are
-independent from one another. Some task ID variables have a more direct
-relationship. One example is `target_date`, which is the date of occurrence of
-the outcome of interest. This task ID is specific for modelling efforts in that
-it is _derived_ from the `origin_date` and `horizon` task ID.
+variables. For example, across 2 `origin_date`s and 3 `location`s, there are 6
+unique combinations of tasks because both of these variables are orthogonal to
+one another. **Derived task ID variables have a direct one to one relationship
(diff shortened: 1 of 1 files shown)


docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,30 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+
+Each model output is defined on the unique combinations across the task ID
+variables. For example, across 10 `origin_date`s and 50 `location`s, there are
+500 unique combinations of tasks because both of these variables are
+independent from one another. Some task ID variables have a more direct
+relationship. One example is `target_date`, which is the date of occurrence of
+the outcome of interest. This task ID is specific for modelling efforts in that
+it is _derived_ from the `origin_date` and `horizon` task ID.
+
+If we had 4 `horizon` values, along with the 10 `origin_dates`, we would have a total of 40 `target_date`s. There are 20,000 _unique_ combinations across 10
```
annakrystalli, line 74: While the walkthrough is useful, I think it would be easier to follow with smaller example numbers. Also the point that needs to be made more clearly is that there is a one to one relationship between a derived task id and the values it depends on. For example a forecast with a given `origin_date` `"2024-11-07"` and horizon `1` has only a single valid `"arget_date` value: `"2024-11-14"`, regardless of the complete list of `target_date` tas ID values.

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,30 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+`origin_date`s, 50 `locations` and the 40 `target_dates`, but there are only
+2,000 _valid and unique_ combinations. While this task ID is useful for modeling
+and visualizations, it must be ignored during validation.
+
+In schemas version 4.0.0, we introduced `derived_task_ids` properties to enable
```
annakrystalli, line 79:
```suggestion
In schema version 4.0.0, we introduced `derived_task_ids` properties to enable
```

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,30 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+hub administrators to define derived task IDs (i.e. task IDs whose values
+depend on the values of other task IDs). The higher level `derived_task_ids
+property` sets the property globally at the hub level but can be overriden by
+the round level `derived_task_ids` property. The property allows for primarily
+validation functionality to ignore such task IDs when appropriate which can
+significantly improve validation efficency. For more information see the
```
annakrystalli, line 85: We should also highlight here in bold that, if any of the task IDs a derived task ID depends on have required values, it is essential for derived_task_ids to be specified. If this is the case and derived task IDs are not specified, the dependent nature of derived task ID values will result in false validation errors when validating required values.

annakrystalli (changes requested):
Thanks for putting this together! Some overall minor comments but just wanted to make sure we clarified that there are situations where derived task IDs **must** be specified for correct validation.

docs/source/user-guide/tasks.md
zkamvar: Done.

docs/source/user-guide/tasks.md
zkamvar: Done in a4ba755f22937ea5763c44853f068070f320e9c6

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,56 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+:::{note}
+If any of the task IDs a derived task ID depends on have `required` values, **it is essential for `derived_task_ids` to be specified**, otherwise, this will result in false validation errors. For example, if you have two required `horizon` values of 1 and 2, then you _must_ include `target_date` in `derived_task_ids`.
+Without specifying this, you will end up with a
+[`req_vals`](https://hubverse-org.github.io/hubValidations/reference/check_tbl_values_required.html)
+check failure. The table below shows an expanded grid of `horizon`, `origin_date`, and `target_date` for a single `origin_date` and two `horizons` with the results of the `req_vals` check if `derived_task_ids` is unspecified (null) or set to `target_date`:
+
+```{table} validation will provide false errors if there are unspecified derived task IDs
+
+| `origin_date` | `horizon` | `target_date` | valid | `derived_task_ids:` `null` | `derived_task_ids:` `["target_date"]` |
+| ----------    | --------- | ------------- | ----------------- | ------------------------ | ----------------------------------- |
```
zkamvar, line 102:
```suggestion
| ------------- | --------- | ------------- | ----- | ------------------------ | ----------------------------------- |
```

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,56 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+:::{note}
+If any of the task IDs a derived task ID depends on have `required` values, **it is essential for `derived_task_ids` to be specified**, otherwise, this will result in false validation errors. For example, if you have two required `horizon` values of 1 and 2, then you _must_ include `target_date` in `derived_task_ids`.
+Without specifying this, you will end up with a
+[`req_vals`](https://hubverse-org.github.io/hubValidations/reference/check_tbl_values_required.html)
+check failure. The table below shows an expanded grid of `horizon`, `origin_date`, and `target_date` for a single `origin_date` and two `horizons` with the results of the `req_vals` check if `derived_task_ids` is unspecified (null) or set to `target_date`:
+
+```{table} validation will provide false errors if there are unspecified derived task IDs
+
+| `origin_date` | `horizon` | `target_date` | valid | `derived_task_ids:` `null` | `derived_task_ids:` `["target_date"]` |
+| ------------- | --------- | ------------- | ----- | ------------------------ | ----------------------------------- |
```
zkamvar, line 102:
```suggestion
| ------------- | --------- | ------------- | ----- | -------------------------- | ------------------------------------- |
```

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,56 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+efforts in that it is _derived_ from the `origin_date` and `horizon` task ID.
+For example, a weekly forecast with a given `origin_date` of "2024-11-07" and
+horizon of 1 has _a single valid_ `target_date` value of "2024-11-14".
+
+If we had 4 `horizon` values, along with the 2 `origin_dates`, we would have a
+total of 8 `target_date`s. There are 192 _unique_ combinations across 4
+`horizon`s, 2 `origin_date`s, 3 `locations` and the 8 `target_dates`, but there
+are only 192/8=24 _valid and unique_ combinations because the 8 `target_date`s
+are _derived_. While this task ID is useful for modeling and visualizations, it
+must be ignored during validation.
```
annakrystalli, line 81:
```suggestion
Each model output task is based on unique combinations of task ID values. For example, for a given `origin_date` which is a task ID which often acts as the round ID (let's say "2024-11-07"), 2 `locations`, and 2 `horizon` values, there are 4 unique tasks (1 `origin_date` × 2 `locations` × 2 `horizon`s). 

**However, some task variables are derived directly from others.** For instance, `target_date` represents when the outcome of interest occurs and is calculated based on the `origin_date` and `horizon`. If the `origin_date` is "2024-11-07" and the `horizon` is 1 week, the `target_date` will be "2024-11-14." For a `horizon` of 2 weeks, the `target_date` will be "2024-11-21." Such task IDs therefore have **a one-to-one relationship** to values of the task IDs they are derived from.

By adding a `target_date` task ID to the above example we would still have a total of 4 unique tasks since `target_date` is derived from the `origin_date` and `horizon` task IDs and each `horizon` produces a unique valid `target_date` per location. 

While derived task IDs like `target_date` are helpful for modeling and visualization, they break assumptions made during some validation tests and can also put significant strain on validation performance. As such they should generally be ignored during standard validation and custom or optional checks added to the validation workflow to check their relationship to the task IDs they are derived from (see for example documentation on the [`opt_check_tbl_horizon_timediff()`](https://hubverse-org.github.io/hubValidations/reference/opt_check_tbl_horizon_timediff.html) used to check the time difference between values in two date columns equals a defined time period defined by values in a horizon column).
```

I've jsut tried to make things simpler and clearer

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,56 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+
+In schema version 4.0.0, we introduced `derived_task_ids` properties to enable
+hub administrators to define derived task IDs (i.e. task IDs whose values
+depend on the values of other task IDs). The higher level `derived_task_ids
```
annakrystalli, line 85:
```suggestion
depend on the values of other task IDs) in their hub config files. The higher level `derived_task_ids
```

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,56 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+| ------------- | --------- | ------------- | ----- | -------------------------- | ------------------------------------- |
+| 2024-11-07 | 1 | 2024-11-14 | 🆗 | ✅ | ✅ |
+| 2024-11-07 | 2 | 2024-11-21 | 🆗 | ✅ | ✅ |
+| 2024-11-07 | 1 | 2024-11-21 | 🚫 | ❌ | NA |
+| 2024-11-07 | 2 | 2024-11-14 | 🚫 | ❌ | NA |
+
+```
+
+As shown in the last column of the above table, the `req_vals` validation passes when `derived_task_ids` is set to `["target_date"]`.
+
```
annakrystalli, line 111: While I like the detail and understand it, I feel it would be useful for someone not steeped in the intricacies of `hubValidations` to confirm that the above detail is helpful rather than confusing.

annakrystalli (changes requested):
Thanks for the additions! I had a go at editing the first  couple of paragraphs to simplify them. Other than that I've made a suggestion that perhaps another hub admin have a look at the added detail (or even all of it for that matter!) to make sure it all makes sense. @LucieContamin perhaps?

docs/source/user-guide/tasks.md
zkamvar: Agreed. I was trying to convey here the specific consequences of not specifying `derived_task_ids`. I think so far we've said "it causes validation errors", but there were no examples that we could point to.

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,47 @@ There can be **one or more task ID variables to define a modeling "target"** (th
 For example, in our [Running Example 1](#running-example-1), the task ID variables are `target`, `location`, and `origin_date`.
 In this example, `target` is the target key and can only take on one value, "inc covid hosp".
 
+#### Derived Task ID variables
+
+Each model output task is based on unique combinations of task ID values. For example, for a given `origin_date` which is a task ID which often acts as the round ID (let's say "2024-11-07"), 2 `locations`, and 2 `horizon` values, there are 4 unique tasks (1 `origin_date` × 2 `locations` × 2 `horizon`s). 
+
+**However, some task variables are derived directly from others.** For instance, `target_date` represents when the outcome of interest occurs and is calculated based on the `origin_date` and `horizon`. If the `origin_date` is "2024-11-07" and the `horizon` is 1 week, the `target_date` will be "2024-11-14." For a `horizon` of 2 weeks, the `target_date` will be "2024-11-21." Such task IDs therefore have **a one-to-one relationship** to values of the task IDs they are derived from.
```
annakrystalli, line 68:
```suggestion
**However, some task ID variables are derived directly from others.** For instance, `target_date` represents when the outcome of interest occurs and is calculated based on the `origin_date` and `horizon`. If the `origin_date` is `"2024-11-07"` and the `horizon` is 1 week, the `target_date` will be `"2024-11-14"`. For a `horizon` of 2 weeks, the `target_date` will be `"2024-11-21"`. Such task IDs therefore have **a one-to-one relationship** to values of the task IDs they are derived from.
```

annakrystalli (approved):
Gah! Found errors in my suggestions 🙈...other than that, we're so close!...maybe it's best to just ship and ask for additional reviews after, maybe in the form of an issue?

zkamvar:
> Gah! Found errors in my suggestions 🙈...other than that, we're so close!...maybe it's best to just ship and ask for additional reviews after, maybe in the form of an issue?

LOL no worries! I think it's okay to wait for @LucieContamin's review since she tagged herself earlier. This topic is also not super critical as we've determined in #193 that only one public hub uses the `derived_task_ids` argument in their workflows.

annakrystalli:
Ok. It's just that the v4 release of all docs hinges on all these being merged but I guess waiting a day would be fine.

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,47 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+property` sets the property globally at the hub level but can be overriden by
+the round level `derived_task_ids` property. The property allows for primarily
+validation functionality to ignore such task IDs when appropriate which can
+significantly improve validation efficency. For more information see the
+[hubValidations documentation on ignoring derived task
+IDs](https://hubverse-org.github.io/hubValidations/articles/validate-pr.html#ignoring-derived-task-ids-to-improve-performance).
+
+:::{note}
+If any of the task IDs a derived task ID depends on have `required` values, **it is essential for `derived_task_ids` to be specified**, otherwise, this will result in false validation errors. For example, if you have two required `horizon` values of 1 and 2, then you _must_ include `target_date` in `derived_task_ids`.
+Without specifying this, you will end up with a
```
LucieContamin, line 86: I struggle with the first sentence a bit, minor issue but I needed it to read it multiple times to understand it, so I might suggest:

```suggestion
If any of the task IDs, which a derived task ID depends on, have `required` values, **it is essential for `derived_task_ids` to be specified**. Otherwise, this will result in false validation errors. For example, if you have two required `horizon` values of 1 and 2, then you _must_ include `target_date` in `derived_task_ids`.
Without specifying this, you will end up with a
```

docs/source/user-guide/tasks.md
LucieContamin: I don't understand the table:
- if the derived task values are correct: the validation passes whether we specify the derived information or not
- if the derived task values are incorrect:
    - the validation fail if we don't provide the derived information
    - the validation returns nothing (ignore) if we do provide the information. 
    
 It seems wrong to me and in contradiction with the "validation will provide false errors if there are unspecified derived task IDs"

LucieContamin (changes requested):
Thank you! I made some minor comments to add more details, or rephrase some sentences a bit, feel free to ignore.
The major issue that I have is the validation example table to me, either I misunderstood something or there is a mistake on that one, would it be possible to add more information or clarify, please?

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,47 @@ There can be **one or more task ID variables to define a modeling "target"** (th
 For example, in our [Running Example 1](#running-example-1), the task ID variables are `target`, `location`, and `origin_date`.
 In this example, `target` is the target key and can only take on one value, "inc covid hosp".
 
+#### Derived Task ID variables
+
+Each model output task is based on unique combinations of task ID values. For example, for a given `origin_date` which is a task ID which often acts as the round ID (let's say "2024-11-07"), 2 `location`s, and 2 `horizon` values, there are 4 unique tasks (1 `origin_date` × 2 `location`s × 2 `horizon`s). 
+
```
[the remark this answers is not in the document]
zkamvar: I think this make sense!

docs/source/user-guide/tasks.md
zkamvar: I don't think it's a minor issue. I also [struggled with the noun phrase "the task IDs a derived task ID depends on"](https://github.com/hubverse-org/hubValidations/pull/161#discussion_r1843974147). I had previously suggested "parent task ID", but [was overruled as not accurate](https://github.com/hubverse-org/hubValidations/pull/161#discussion_r1843996007). I would like to come up with a succinct descriptor of the relationship so that we can clearly label "task IDs a derived task ID depends on."

docs/source/user-guide/tasks.md
zkamvar: @LucieContamin this is exactly the kind of feedback I was looking for! The way it works for the validation is that the derived task IDs are omitted from the validation table when `derived_task_ids` is specified. 

I think it's probably best to remove the table or split it into two tables. I'll make a change and see if it works.

docs/source/user-guide/tasks.md
zkamvar: Based on @LucieContamin's suggestion in slack: 

```suggestion
If any task IDs with `required` values have an associated derived task ID, **it is essential for `derived_task_ids` to be specified**. Otherwise, this will result in false validation errors. For example, if you have two required `horizon` values of 1 and 2, then you _must_ include `target_date` in `derived_task_ids`.
Without specifying this, you will end up with a
```

docs/source/user-guide/tasks.md
```diff
@@ -61,6 +61,47 @@ There can be **one or more task ID variables to define a modeling "target"** (th
...
+**However, some task ID variables are derived directly from others.** For instance, `target_date` represents when the outcome of interest occurs and is calculated based on the `origin_date` and `horizon`. If the `origin_date` is `"2024-11-07"` and the `horizon` is 1 week, the `target_date` will be `"2024-11-14"`. For a `horizon` of 2 weeks, the `target_date` will be `"2024-11-21"`. Such task IDs therefore have **a one-to-one relationship** to values of the task IDs they are derived from.
```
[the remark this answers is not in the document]
zkamvar: Thank you, @LucieContamin! I've cleaned up your suggestion a bit below:

```suggestion
**However, it is possible to have task ID variables that are derived directly from others.** For example, `target_date` (which represents when the outcome of interest occurs) can be calculated based on the `origin_date` and `horizon` (e.g. `origin_date + horizon * 7` to calculate weekly predictions). If the `origin_date` is `"2024-11-07"` and the `horizon` is 1 week, the `target_date` will be `"2024-11-14"`. For a `horizon` of 2 weeks, the `target_date` will be `"2024-11-21"`. Such task IDs therefore have **a one-to-one relationship** to values of the task IDs they are derived from. We strongly advise hub administrators to add the derived information and calculation in their documentation. 
```

docs/source/user-guide/tasks.md
zkamvar: I attempted a fix for this in 4cd5a567ceecb5e1b1b692a97d81136c65e9f077

PR #220, opened 2024-12-05
add documentation about disallowing optional quantiles

### LINKED ISSUE #192: Add documentation of why we disallow optional quantiles
Regarding changes in https://github.com/hubverse-org/schemas/issues/102 and https://github.com/orgs/hubverse-org/discussions/24

From @harryhoch: It seems as if some documentation should be developed describing why optional quantiles are not a good idea.

The tasks:

 - [x] figure out what the actual advice should be
 - [x] identify target audience
 - [x] identify places in documentation it should be referred to

zkamvar:
I added the discussion in the Defining modeling tasks chapter and I also provided an example of optional task ID variables because it helped set the contrast. 

I also included example JSON because it's often helpful to have a visual aid. 

This will fix #192

### DIFF
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -49,6 +49,41 @@ Moreover, these tables are presented in a [long/narrow representation](https://e
 
 [^tidy]: This type of data is also known as "tidy data," a term coined by Hadley Wickham that's heavily used in the R community. You can read more about the concept in the [Data tidying chapter of the R4DS book](https://r4ds.hadley.nz/data-tidy#sec-tidy-data) and the [Tidy Data paper by Wickham (2014)](https://www.jstatsoft.org/article/view/v059i10).
 
+In the `tasks.json` file, task ID variables are a collection of JSON objects
+that define required and optional values for these variables. In the example below from [the COVID-19 variant nowcast hub](https://github.com/reichlab/variant-nowcast-hub/blob/main/hub-config/tasks.json), there are four task ID variables defined: `"nowcast_date"`, `"target_date"`, `"location"`, and `"clade"`. 
+
+```json
+"task_ids": {
+    "nowcast_date": {
+        "required": [
+            "2024-09-11"
+        ],
+        "optional": null
+    },
+    "target_date": {
+        "required": null,
+        "optional": ["2024-08-11", "2024-08-12", "2024-08-13", "2024-08-14", "2024-08-15", "2024-08-16", "2024-08-17", "2024-08-18", "2024-08-19", "2024-08-20", "2024-08-21", "2024-08-22", "2024-08-23", "2024-08-24", "2024-08-25", "2024-08-26", "2024-08-27", "2024-08-28", "2024-08-29", "2024-08-30", "2024-08-31", "2024-09-01", "2024-09-02", "2024-09-03", "2024-09-04", "2024-09-05", "2024-09-06", "2024-09-07", "2024-09-08", "2024-09-09", "2024-09-10", "2024-09-11", "2024-09-12", "2024-09-13", "2024-09-14", "2024-09-15", "2024-09-16", "2024-09-17", "2024-09-18", "2024-09-19", "2024-09-20", "2024-09-21"]
+    },
+    "location": {
+        "required": null,
+        "optional": ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY", "PR"]
+    },
+    "clade": {
+        "required": ["24A", "24B", "24C", "recombinant", "other"],
+        "optional": null
+    }
+}
+```
+
+In this particular round, modelers MUST submit predictions with the
+`"nowcast_date"` of 2024-09-11 with all of the `"clade"`s (24A, 24B, 24C,
+recombinant, and other). Any submissions that contain anything other than those
+exact values will result in an error. In contrast, modellers MAY submit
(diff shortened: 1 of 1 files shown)


docs/source/user-guide/tasks.md
```diff
@@ -49,6 +49,41 @@ Moreover, these tables are presented in a [long/narrow representation](https://e
...
 
+In the `tasks.json` file, task ID variables are a collection of JSON objects
+that define required and optional values for these variables. In the example below from [the COVID-19 variant nowcast hub](https://github.com/reichlab/variant-nowcast-hub/blob/main/hub-config/tasks.json), there are four task ID variables defined: `"nowcast_date"`, `"target_date"`, `"location"`, and `"clade"`. 
+
+```json
+"task_ids": {
+    "nowcast_date": {
+        "required": [
+            "2024-09-11"
+        ],
```
LucieContamin, line 60:
```suggestion
        "required": ["2024-09-11"],
```
Very minor update, just to keep the same style

docs/source/user-guide/tasks.md
```diff
@@ -49,6 +49,41 @@ Moreover, these tables are presented in a [long/narrow representation](https://e
...
+```
+
+In this particular round, modelers MUST submit predictions with the
+`"nowcast_date"` of 2024-09-11 with all of the `"clade"`s (24A, 24B, 24C,
+recombinant, and other). Any submissions that contain anything other than those
+exact values will result in an error. In contrast, modellers MAY submit
+predictions for ANY of the `"target_date"`s between 2024-08-11 and 2024-09-21
+and ANY of the states listed in `"location"`. By allowing modelers to submit a
+subset of optional values, it means poor-performing models can be omitted so
+they do not negatively influence model ensembles.
```
LucieContamin, line 85:
```suggestion
and ANY of the states listed in `"location"`. By allowing modelers to submit a
subset of optional values, it means poor-performing models can be omitted so
they do not negatively influence model ensembles. It also allows models that 
have only have the capacity for a subset of the optional options can still
participate. 
```
I add a sentence for why we also have optional values in SMH (to allow team to only subset some states for example, if they don't have the capacity to run national). I will ask someone from SMH to validate.

docs/source/user-guide/tasks.md
```diff
@@ -89,9 +124,72 @@ In those cases, the new variables should be added to this list to ensure that th
 (output-types)=
 ## Output types
 
-The `output_type` object defines accepted representations for each task.
+The `output_type` object defines accepted model output representations for each task. These define what kind of model output is expected, what range of values
+we expect, if multiple values are expected, what identifies that value, and
+whether or not the output type is required for submission.
```
LucieContamin, line 129:
```suggestion
we expect, if multiple values are expected, what identifies those values, and
whether or not if each output type is required for submission.
```
Not sure it's correct (English is not my first language!), but I update a bit the phrase, feel free to ignore. 

I am not sure I understand what "multiple values" means here, we always have multiple value, no?

docs/source/user-guide/tasks.md
```diff
@@ -89,9 +124,72 @@ In those cases, the new variables should be added to this list to ensure that th
...
+        "minimum": 0
+    },
+    "is_required": true
+}
+```
+
+From the code block above, you can see that an output type has four components:
+
+1. (line 1) `"quantile"`{.codeitem} the name of the output type representation
+   (e.g. `"mean"`, `"quantile"`, `"pmf"`)
```
LucieContamin, line 157:
```suggestion
   (e.g. "cdf"`, `"mean"`, "median"`, `"quantile"`, `"pmf"`, `"sample"`) 
```

docs/source/user-guide/tasks.md
```diff
@@ -89,9 +124,72 @@ In those cases, the new variables should be added to this list to ensure that th
...
+    "is_required": true
+}
+```
+
+From the code block above, you can see that an output type has four components:
+
+1. (line 1) `"quantile"`{.codeitem} the name of the output type representation
+   (e.g. `"mean"`, `"quantile"`, `"pmf"`) 
+2. (line 2) `"output_type_id"`{.codeitem} In the case of quantiles, the output
+   type ID is an indcation of the quantile bins. **Unlike task IDs, all
```
LucieContamin, line 159:
```suggestion
   type ID is an indication of the quantile bins. **Unlike task IDs, all
```

LucieContamin (commented):
Thank you very much! I made some minor comment, but I mainly had a question I wanted to clarify before approving if that's ok.

docs/source/user-guide/tasks.md
zkamvar: Good catch! English _is_ my first language and my draft was definitely wrong 🙈

Hopefully this fixes it: 

```suggestion
we expect, if multiple values are expected, what identifies those values (e.g. a bin, category, or ID), and
whether or not the output type is required for submission.
```

> I am not sure I understand what "multiple values" means here, we always have multiple value, no?

point estimates will only have a single value per task ID.

docs/source/user-guide/tasks.md
zkamvar: Thank you!

docs/source/user-guide/tasks.md
LucieContamin: Ah I see, thank you. I was not reading it like this. It makes more sense now. thanks again fro the clarification

docs/source/user-guide/tasks.md
zkamvar: Hi @LucieContamin, have you confirmed that this is accurate?

LucieContamin: Yes sorry that is correct

LucieContamin (approved):
Thank you! It looks all good to me. I just added 2 very minor suggestions.

docs/source/user-guide/tasks.md
```diff
@@ -89,9 +124,72 @@ In those cases, the new variables should be added to this list to ensure that th
...
+        "minimum": 0
+    },
+    "is_required": true
+}
+```
+
+From the code block above, you can see that an output type has four components:
+
+1. (line 1) `"quantile"`{.codeitem} the name of the output type representation
+   (e.g. "cdf"`, `"mean"`, "median"`, `"quantile"`, `"pmf"`, `"sample"`)
```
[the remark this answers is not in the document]
zkamvar: 🙈 mea culpa. I should have caught those

PR #222, opened 2024-12-10
Minor clarifications on derived and standard task IDs

annakrystalli:
This PR adds clarification to the situations when specifying derived task IDs is necessary for correct validation.

I also noticed some areas in the standard task IDs section that I feel could be clearer. 
1. I've brought forward the explanation of the reasoning behind standard task IDs to the intro of the section.
2. Changed the title which seemed confusing to me as it was


zkamvar (approved):
Thank you! I agree that this makes the relationship more clear. Thank you also for removing the "Proposed" from that title and moving the rationale up.

LucieContamin (approved):
Looks good to me too!

annakrystalli:
Thanks both!

PR #228, opened 2024-12-24
fix syntax highlighting for json code block

zkamvar:
sphinx tries its best to highlight code blocks, but sometimes it just can't because it cannot tolerate any mistakes in the syntax. Making the code block an explicit `{code-block}` class and adding `:force: true` to the directives tells sphinx to highlight the best it can because the world's not perfect.


zkamvar:
I'm going to merge this in because it's a minor formatting change that does not affect content.

PR #229, opened 2025-01-03
replace code portion of scripting chapter with link

zkamvar:
The code in this chapter had previously produced an invalid configuration file. By linking to [the scripting vignette](https://hubverse-org.github.io/hubAdmin/dev/articles/scripting-tasks-config.html), we can make sure the code is always correct and up to date with current practices. 

Note that the vignette itself is derived from this chapter and there is still room for improvement (see https://github.com/hubverse-org/hubAdmin/issues/91)


zkamvar:
thank you!

PR #235, opened 2025-01-17
Use uv for hubDocs development setup

### LINKED ISSUE #174: remove numpy from this project
I am having a bear of a time trying to get things installed on my Mac and one thing that's causing me a headache right now is numpy, which was introduced in https://github.com/hubverse-org/hubDocs/commit/ebc22d66a26dd0f80e4ded25b43485bc6d7c35bd.... I don't think we actually use numpy anywhere. Could we remove it?

bsweger:
Closes #174 

**Background**

This PR tackles the issue above (removing numpy) by updating the Python-based development workflow instructions.

The changes are an experiment in (yet another, I know) workflow for our Python projects, and I tried to strike a balance between reducing project setup burden for non-Python folks and providing flexibility for contributors who already have a preferred set of Python tools.

**Notes for reviewers**

- The changes can be reviewed commit-by-commit
- Please try out the new site preview instructions on the README (feedback from Windows users is especially needed)


pyproject.toml
```diff
@@ -1,8 +1,46 @@
+[project]
+name = "hubdocs"
+description = "Documentation for the Hubverse collection of collaborative modeling tools"
+authors = [
+    {name = "Consortium of Infectious Disease Modeling Hubs"},
+    {name = "Graziella", email = "graziella@lumache.com"}
```
zkamvar, line 6: I think Graziella was from the readthedocs template. I don't know if she is real. 

```suggestion
    {name = "Consortium of Infectious Disease Modeling Hubs"}
```

zkamvar (approved):
Thank you! This is absolutely a major improvement. 

I was actually using the makefile for development, but my process was not great (context: I work directly from the terminal)

```
source .venv/bin/activate
cd docs/
nvim source/path/to/file.md
:make html
(in another window)
python -m http.server 8080 -d build/html
```

That being said, the `uv run sphinx-autobuild docs/source docs/_build/html` is a bit of a mouthful (and begins to reminds me of the [fiasco that is building with Jekyll](https://github.com/carpentries/styles/blob/7bcdf5091ceb681ea3a239c83f3351fde09c1580/Makefile#L6)) and would appreciate a similar makefile at the top so that I could do the following:

```
(in another window)
make watch
(in editor window)
nvim docs/source/path/to/file.md
```

...but that's out of scope of this PR and I approve.

pyproject.toml
bsweger: I was wondering who that was 😂

README.md
```diff
@@ -1,98 +1,118 @@
...
-conda activate hubdocs
-```
+    ```script
+    build succeeded, 4 warnings.
 
-### Deactivate environment
+    The HTML pages are in docs/_build/html.
+    [sphinx-autobuild] Serving on http://127.0.0.1:8000
+    [sphinx-autobuild] Waiting to detect changes...
+    ```
```
micokoch, line 80: So far, this is great and easy to use. I would add a few steps that may be obvious to programmers, but not others:
- If you click on the link (http://127.0.0.1:8000/) you can see the built page.
- The build is running until you press CTL-C to exit.
- Maybe how to leave the virtual environment with `deactivate`

micokoch (changes requested):
Hi Becky, this is great and it worked well when I tested it. I didn't test adding dependencies, but on a superficial review, it seemed to work well. I made three small suggestions of steps you could add that may be obvious to you but weren't to me when I first built the site.

README.md
```diff
@@ -1,98 +1,118 @@
...
-### Create environment and install dependencies
+1. Install uv on your machine (you will only need to do this once):
+<https://docs.astral.sh/uv/getting-started/installation/>
+2. Clone this repository. The rest of the instructions should be executed from
+the repo's root directory.
+3. Create a virtual environment for the project:
 
-You first need to create a python 3.9  `hubdocs` environment and install all additonal python dependencies within it.
+    ```script
+    uv venv --seed
```
annakrystalli, line 55: I'm a little confused as to whether creating the environment also activates it? When I ran:
```
uv venv --seed
```
I got 
```
Using CPython 3.9.15 interpreter at: /Users/Anna/opt/anaconda3/envs/hubdocs/bin/python
Creating virtual environment with seed packages at: .venv
 + pip==24.3.1
 + setuptools==75.8.0
 + wheel==0.45.1
Activate with: source .venv/bin/activate
```
So I'm not sure if I need to run `Activate with: source .venv/bin/activate` or not?

Might be good to clarify this in this section either way

README.md
```diff
@@ -1,98 +1,118 @@
...
 
-```bash
-# Create hubdocs environment containing python 3.9
-conda create -n hubdocs python=3.9
+4. Install dependencies:
 
-# Activate hubdocs environment
-conda activate hubdocs
+    ```script
+    uv pip install -r requirements/requirements.txt
```
annakrystalli, line 61: When I try to run this, I get 
```
uv pip install -r requirements/requirements.txt
  × No solution found when resolving dependencies:
  ╰─▶ Because the current Python version (3.9.15) does not satisfy Python>=3.10 and alabaster==1.0.0
      depends on Python>=3.10, we can conclude that alabaster==1.0.0 cannot be used.
      And because you require alabaster==1.0.0, we can conclude that your requirements are unsatisfiable.
```
In my above comment it looks like `uv` is using `Using CPython 3.9.15 interpreter at: /Users/Anna/opt/anaconda3/envs/hubdocs/bin/python`. I think there might need to be more details of how/where to install the appropriate version of python and set `uv` to use it?

Am I right in understanding that with `uv`, the environment does not contain a specific version of python? This seems to be a difference compared to the conda workflow that users that don't use python that often might need some help with.

annakrystalli (commented):
I tried to run the workflow but got a bit confused by the python version `uv` was trying to use and ultimately got stuck on installing the dependencies. As someone who rarely uses python (and find the multiple versions and location python can end up on my machine quite confusing), this could be an area that the instructions could benefit with some additional detail.

README.md
bsweger: Oh, interesting test case--thanks for testing this out!

From your prior note:

> So I'm not sure if I need to run Activate with: source .venv/bin/activate or not?

One of the neat things about uv is that you can let it handle virtual environments "behind the scenes" and not worry about it too much:

- uv _does_ install Python when it creates project-based virtual environments: https://docs.astral.sh/uv/concepts/python-versions/#project-python-versions
- After you create a virtual env via the `uv venv --seed` command, subsequent commands prefixed by `uv` (e.g., `uv pip install`) are smart enough to see the virtual environment and just use it (no explicit activation required)

However, if there's _already_ an activated virtual environment, uv will respect that and use it: https://docs.astral.sh/uv/pip/environments/#discovery-of-python-environments

> When running a command that mutates an environment such as uv pip sync or uv pip install, uv will search for a virtual environment in the following order:
> - An activated virtual environment based on the VIRTUAL_ENV environment variable.
> - An activated Conda environment based on the CONDA_PREFIX environment variable.
> - A virtual environment at .venv in the current directory, or in the nearest parent directory.

Based on the error message above, I'm guessing you might have an active conda envrionment?

If that's true, can you try explicitly activating the virtual environment via the `source .venv/bin/activate` incantation (or deactivating CONDA) and then trying `uv pip install -r requirements/requirements.txt` again?

If we can confirm that this is, in fact, the issue, I'll update the instructions to clarify.

annakrystalli: Thanks for the tips!

I reckon it's indeed sth to do with my conda setup.
So I've not activated the environment and I get:
```
(base) ☁  hubDocs [bs/update-python-setup/174] uv venv --seed                                 
Using CPython 3.9.12 interpreter at: /Users/Anna/opt/anaconda3/bin/python3
Creating virtual environment with seed packages at: .venv
 + pip==24.3.1
 + setuptools==75.8.0
 + wheel==0.45.1
Activate with: source .venv/bin/activate
```
it's trying to use my conda python which is older and hence the installation fails because it needs python v3.10 or above

but activating the environment also doesn't help as it's trying to use the same conda version
```
(base) ☁  hubDocs [bs/update-python-setup/174] source .venv/bin/activate
(hubDocs) (base) ☁  hubDocs [bs/update-python-setup/174] uv pip install -r requirements/requirements.txt
  × No solution found when resolving dependencies:
  ╰─▶ Because the current Python version (3.9.12) does not satisfy Python>=3.10 and alabaster==1.0.0
      depends on Python>=3.10, we can conclude that alabaster==1.0.0 cannot be used.
      And because you require alabaster==1.0.0, we can conclude that your requirements are unsatisfiable.
(hubDocs) (base) ☁  hubDocs [bs/update-python-setup/174] 
```
I've probably set this somewhere and don't remember where 😂 sorry I know i could get to the bottom of this but thought I'd get this in the open as maybe other more sporadic users might also trip up? Or maybe it's just me!

annakrystalli: Thanks for helping me troubleshoot!  For transparency, I was in a conda activated environment when I ran `uv venv --seed` so it was using the older python version in the environment. Deactivating the environment with `conda deactivate` (not `deactivate` which I tried first!) fixed the problem and I was able to install everything successfully 🎉

README.md
bsweger: Thanks for trying this out! I pushed some changes to address your suggestions, including a note that you don't have to activate the virtual environment...does that address your third bullet about how to leave the environment?

annakrystalli (approved):
Thanks for adding the warning about conda environments! Everything works now and was able to very easily build and preview the site! A great improvement in workflow 🤩

lshandross (approved):

README.md
micokoch: Great, thanks Becky!

micokoch (approved):

README.md
```diff
@@ -1,98 +1,131 @@
...
+long as dependency updates follow the workflow of adding (or removing)
+dependencies from `pyproject.toml` and re-generating an annotated
+`requirements/requirements.txt` file. In other words, don't add update
+the requirements.txt file directly.
 
-### Using a conda environment for local development
+> [!IMPORTANT]
+> If you have an active Python virtual environment (for example, conda's
+> base environment), you'll need to deactivate it before following the
+> instructions below.
```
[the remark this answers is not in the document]
bsweger: Thanks for giving this a try--very happy we have a Windows perspective represented (and very happy that uv installed on your machine!)

Do you by any chance have the output of your `uv venv --seed` command available? If so, could you share it? I want to make sure `uv` didn't actually use your base conda's Python interpreter when it created the virtual environment.

I'm inclined to leave the wording because it's possible that someone's base conda uses a version of Python that's compatible with hubDocs, so the instructions here would work. Even so, we probably don't want the project's virtual environment tied to a conda-managed Python if we can help it.

bsweger:
> That being said, the `uv run sphinx-autobuild docs/source docs/_build/html` is a bit of a mouthful (and begins to reminds me of the [fiasco that is building with Jekyll](https://github.com/carpentries/styles/blob/7bcdf5091ceb681ea3a239c83f3351fde09c1580/Makefile#L6)) and would appreciate a similar makefile at the top so that I could do the following:
> 
> ```
> (in another window)
> make watch
> (in editor window)
> nvim docs/source/path/to/file.md
> ```
> 
> ...but that's out of scope of this PR and I approve.

@zkamvar `sphinx-autobuild [input] [output]` is the standard autobuild command for a typically-configured sphinx project, and it spins up a server and hot-reloads as changes are made. I'm not sure I follow what you'd like a makefile to do--can you explain more (agree that it's out of scope here, but we can file an issue)

zkamvar:
> I'm not sure I follow what you'd like a makefile to do--can you explain more (agree that it's out of scope here, but we can file an issue)

For context: my workflow was informed by the structure presented in the [minimal sphinx example from readthedocs](https://docs.readthedocs.io/en/stable/examples.html#sphinx) that shows [the usage of the Makefile they embedded in the project](https://github.com/readthedocs-examples/example-sphinx-basic/?tab=readme-ov-file#example-project-usage). Note also this is where "Lumanche" came from. I just find it easier to type `make <thing>` than to have to remember four different concepts (uv, sphinx, input, and output) in the `uv run sphinx-autobuild [input] [output]` command. 

That all being said, I only know about sphinx via readthedocs, and I suspect their examples are a bit contrived, so I'm happy to go with the idiomatic sphinx (which is an improvement!)

README.md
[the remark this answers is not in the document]
lshandross: Here's the output:

```
(base) C:\Users\lshandross\Documents\hubDocs>uv venv --seed
Using CPython 3.12.7 interpreter at: C:\Users\lshandross\AppData\Local\anaconda3\python.exe
Creating virtual environment with seed packages at: .venv
 + pip==24.3.1
Activate with: .venv\Scripts\activate
```

bsweger:
> > I'm not sure I follow what you'd like a makefile to do--can you explain more (agree that it's out of scope here, but we can file an issue)
> 
> For context: my workflow was informed by the structure presented in the [minimal sphinx example from readthedocs](https://docs.readthedocs.io/en/stable/examples.html#sphinx) that shows [the usage of the Makefile they embedded in the project](https://github.com/readthedocs-examples/example-sphinx-basic/?tab=readme-ov-file#example-project-usage). Note also this is where "Lumanche" came from. I just find it easier to type `make <thing>` than to have to remember four different concepts (uv, sphinx, input, and output) in the `uv run sphinx-autobuild [input] [output]` command.
> 
> That all being said, I only know about sphinx via readthedocs, and I suspect their examples are a bit contrived, so I'm happy to go with the idiomatic sphinx (which is an improvement!)

Thanks for the detail--I see what you're saying about the overloaded command. If it turns out to be a hassle, file an issue, and we shall make a new make.

PR #239, opened 2025-01-28
update readme

### LINKED ISSUE #205: Update and simplify README installation instructions
At the moment, the readme is not quite accurate:

 - We do not use a Jupyter book theme, we use the sphinx book theme
 - Installation individual packages to the global python environment is frowned upon in the Python world (so I understand). We should provide instructions to set up a virtual environment
 - Since it ensh\*ttified, we should no longer recommend `conda` for setting up python. We should probably just move to recommending `uv` (but maybe this is to @bsweger's chagrin).

micokoch:
Closes #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,11 @@
-# Documentation for Infectious Disease Modeling Hubs
+# Documentation for Collaborative Modeling Hubs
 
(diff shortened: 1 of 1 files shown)


zkamvar (approved):
LGTM, thank you!

PR #246, opened 2025-02-06
improving contributors page

micokoch:
Improvements to contributor page (and script that builds it) + some linting.


micokoch:
Closes #216

scripts/contributors.py
```diff
@@ -0,0 +1,132 @@
...
+contributors = defaultdict(lambda: defaultdict(int))
+
+# Dictionary to store contributor commit counts and last contribution date
+contributors = defaultdict(lambda: {"repos": defaultdict(int), "last_commit": None})
+
+# Fetch contributors for each repository
+for repo in repos:
+    url = f"{base_url}/repos/{org}/{repo}/contributors"
+    response = requests.get(url, headers=headers)
+    if response.status_code == 200:
```
zkamvar, line 40: Instead of

```
if response.status_code == 200: 
     # do thing
     # that takes
     # multiple lines
```

You should use a guard clause, which avoids indentation and, importantly, frees up cognitive load because the functionality is on the happy path of the code. 

```
if response.status_code != 200:
    continue
    # message or error that something is not okay
# do thing
# that takes
# multiple lines
```

zkamvar: this strategy applies to all the `if response.status_code == 200` conditionals below.

zkamvar (changes requested):
Thank you, @micokoch! I think this is a good first stab and I have some suggestions. Mainly, it's not a good idea to sort and show commit counts. I think the idea of showing what repositories were contributed to makes sense, but it would be better to simply list them in alphabetical order. 

The reason why commit counts is not good is partially because number of commits do not necessarily translate to any sort of indication of amount of code or quality (i.e. https://github.com/hubverse-org/hub-dashboard-control-room/pull/25/commits/987d89457c96192108f68b0f73e169b965ac0049 is one of 60 commits and counting...many of which are simply testing minor things in GitHub workflows) and will always highlight folks who are able to regularly contribute as part of their jobs. 

Instead, you could show a person's profile with links to the repositories they contributed to below.

micokoch:
Thanks @zkamvar and @bsweger - I made the changes you requested. If it looks okay to both of you, please approve. Thanks!

bsweger (approved):
Thanks for iterating on the contributors page!

zkamvar (approved):
Looks good! Only one suggestion to clearly indicate that the order is random.

PR #251, opened 2025-02-17
Add description of null target_keys. Resolves #250

### LINKED ISSUE #250: Add description of null target keys
Issue on page /user-guide/tasks.html

The `target_keys` definition needs expanding to include the option to use null values.

From The description of the property in the schema:

> "Should be either null, in the case where the target is not specified as a task_id and is specified solely through the target_id target_metadata property or an object with one property, the name of which must match a single task_id variable named within the same model_tasks object. The property should have one specified task ID value. Each value defines a single target value."

annakrystalli:
This PR adds a description of the option and situations in which `null`  `target_keys` can be used.

Resolves #250


elray1 (approved):
lgtm!

Issue #255, opened 2025-03-18
MyST: Missing cross references and other warnings


zkamvar:
As I'm currently working on #253, I'm finding a lot of warnings pop up that have to do with missing cross references (under `docs/source`), which means that if someone clicks on any of these links, they will probably go nowhere (which Arnie is fine with, but Gilbert is gonna get all upset and moody).

| path | type | info |
| ---  | ---  | ---  |
| overview/contributors.md:123: | ERROR | Document may not end with a transition. |
| overview/who-uses-hubverse.md:9: | WARNING | 'image': Unknown option keys: ['caption'] (allowed: ['align', 'alt', 'class', 'height', 'loading', 'name', 'scale', 'target', 'width']) [myst.directive_option] |
| user-guide/tasks.md:34: | WARNING | duplicate label hub-data-relations, other instance in user-guide/dashboards.md |
| developer/python.md:6: | WARNING | 'checklists' [myst.xref_missing] |
| developer/python.md:7: | WARNING | 'development-checklist' [myst.xref_missing] |
| developer/python.md:8: | WARNING | 'release-checklist' [myst.xref_missing] |
| developer/python.md:9: | WARNING | 'hotfix-checklist' [myst.xref_missing] |
| developer/python.md:10: | WARNING | 'releasing-python-packages' [myst.xref_missing] |
| developer/python.md:11: | WARNING | 'overview' [myst.xref_missing] |
| developer/python.md:12: | WARNING | 'testpypi-setup' [myst.xref_missing] |
| developer/python.md:13: | WARNING | 'pypi-setup' [myst.xref_missing] |
| developer/python.md:14: | WARNING | 'add-package-maintainers' [myst.xref_missing] |
| developer/python.md:16: | WARNING | 'releasing-python-packages' [myst.xref_missing] |
| developer/python.md:58: | WARNING | local id not found in doc 'developer/release-process': 'releases' [myst.xref_missing] |
| developer/python.md:80: | WARNING | local id not found in doc 'developer/release-process': 'patch' [myst.xref_missing] |
| developer/release-process.md:105: | WARNING | 'previous-state' [myst.xref_missing] |
| developer/release-process.md:143: | WARNING | 'new-release-process' [myst.xref_missing] |
| developer/release-process.md:143: | WARNING | 'dev' [myst.xref_missing] |
| overview/who-uses-hubverse.md:46: | WARNING | 'model-output-format' [myst.xref_missing] |
[...]

PR #267, opened 2025-03-19
Fix contributors transition error and homogenize last two spaces

micokoch:
When complete, fixes issue #255


micokoch:
As far as I tested, this should solve #255 . Thanks for reviewing @zkamvar !

docs/source/developer/release-process.md
```diff
@@ -196,6 +198,7 @@ gitGraph
   `expand_model_out_grid()`. Users who change nothing in their workflows or
   scripts will not see any change for the better or worse.
 
+(patch)=
 <a id="patch">`Patch`</a>
```
zkamvar, line 202: This is the correct pattern to use. I originally had the anchor link to get around my ignorance for how to add span links according to https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html

In addition, this removes the `<a>` tag that I inserted in a desperate attempt to get an ID. 

Note that this should be applied to the other elements in this definition list. I cannot make suggestions for those here because they are not diffs.

```suggestion
{#patch}
`Patch`
```

docs/source/developer/release-process.md
```diff
@@ -206,6 +209,7 @@ gitGraph
   so [version 1.0.1](https://hubverse-org.github.io/hubAdmin/news/index.html#hubadmin-101)
   was a patch release.
 
+(dev)=
 <a id="dev">dev versioning</a>
```
zkamvar, line 213:
```suggestion
{#dev}
dev versioning
```

zkamvar (approved):
Thank you for this! 

There's only a couple of minor changes that I highlighted, but they are minor enough that they can be taken care of in a separate PR. 

Good work!

PR #287, opened 2025-04-01
Tell dependabot to consolidate minor dependency updates into a single

bsweger:
To reduce the amount of "noise" from Dependabot-submitted pull requests, this changeset adds a "groups" setting to dependabot.yaml. More information about the groups setting:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

The changeset also changes the "pip" package-ecosystem directory from "requirements" to the repo's root, where pyproject.toml is located.


zkamvar (approved):
LGTM!

PR #304, opened 2025-04-23
znk/target data updates

### LINKED ISSUE #297: how to write target-data information to add
I apologize if I miss it in the current documentation, should we add documentation for the hub admin on how to write/store the target-data data. For example, that it should be in an unique or partitioned files in csv, parquet format with a standard file name and folder name, etc.

I can not find the information in the current doc and as we are releasing validation to test it, we should document it.

zkamvar:
This will fix #297 

- **fix broken reference to dashboards**
- **add anchors to the hub structure page**
- **add context for target data file conventions**

### DIFF
--- a/docs/source/user-guide/hub-structure.md
+++ b/docs/source/user-guide/hub-structure.md
@@ -12,12 +12,14 @@ The directory and file structure of a modeling hub should contain only the follo
 
 ## Required Components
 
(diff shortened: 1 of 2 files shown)


LucieContamin (approved):
it looks great, thank you very much!

PR #305, opened 2025-04-24
Update the cloud onboarding section of Hubverse dev docs

bsweger:
This PR adds a section about offboarding hubs (mostly as a reminder that the contents of an S3 bucket need to be removed before the bucket itself can be deleted)


annakrystalli (approved):
Looks great! Thanks for the updates!

Issue #211, opened 2024-11-13
Revamp documentation structure for discoverability and targeting


zkamvar:
Given the experience of submissions to https://github.com/reichlab/variant-nowcast-hub, one of the main concerns is that it was not clear what packages modelers should use to support their submissions, partially because the documentation was not clear.

This highlights a need for broad restructuring of the documentation. Discussion from the 2024-11-13 meeting are as follows:

> Zhian: Our main place for docs right now doesn’ evaluate code
> Anna:
> Could have a section of documentation that is a quickstart specifically for each user role (admin exists now, but what about toher groups like modelers, analysts)?
> Package articles and vignettes
> Readmes for packages, link out to specific package from main hubDocs
> Harry: It might also help to have a super high-level introduction that lays out the broad strokes of how hubs work and the different roles that participants might play in hubs. This will help people learn which “getting started” spot would be most appropriate at any given time.
> Zhian: Melissa’s diagram could inform this
> Anna: could also pull from Kath’s higher level figure
> There is a version of this in the manuscript; Nick will file an issue to add
> Nick Reich (Nov 13, 2024, 9:28 AM): I agree! We’ve tried to do this here: https://hubverse.io/en/latest/overview/who-uses-hubverse.html 
> But we haven’t really created the detailed “Quickstart” guide for each kind of user.
> Hochheiser, Harry (Nov 13, 2024, 9:29 AM): This is good. A figure or two might help make it even better.  It also might help to add a definition of a hub and a summary of the workflow to the intro page.
> Becky: We have heard feedback from modelers that it’s hard to know where to get started
> Evan:
> There is a vignette in the hubverse R package
> Any documentation solution should be able to describe packages in multiple languages
> Becky: also need something that says how to onboard a cloud hub (I lost track of this)
> Nick: the menu on the left of the docs is too long.  Could we collapse some subcategories?  This would also give us room to add more headers to “external documentation”.
> Harry: language-specific orientation might help. “If you are an R  person, go here… If you are a Python person go here…”
> Harry - consider moving away from the term “stakeholder”. See - https://doi.org/10.1007/s11625-024-01496-4, for example.
> Zhian has volunteered to put together a plan for documentation.  (not highest/immediate priority, but soon)
>

zkamvar:
I believe this is _largely_ complete with the release of hubverse.io (built from https://github.com/hubverse-org/hubverse-site).

[...]

PR #314, opened 2025-05-02
Restructuring hubDocs

micokoch:
Now that we have such a nice website, this is my first pass at restructuring the website to make it closer to a documentation site. I haven't deleted duplicate pages, but I did reorganize it. Does the structure seem okay to @nickreich @annakrystalli @zkamvar ? Please let me know what you think.


micokoch:
Thanks @annakrystalli ! You make very good points. I threw this out more as a proposal, and if it doesn't make sense, I can be more careful and gradual about any changes and do it in other separate PRs (and just discard this one). But, if you do have time to make specific recommendations, I'd love to hear them.

docs/source/index.md
```diff
@@ -1,17 +1,25 @@
 # The hubverse: open tools for collaborative modeling
 
+```{caution}
+:class: warning
+
+🚧 This site is being restructured
+
+The main hubverse landing page has migrated to [https://hubverse.io/](https://hubverse.io/). This site is being restructured to contain more detailed information on setting up a modeling hub.
+```
+
```
zkamvar, line 10: I would keep this, but not make it so scary. 


````suggestion
```{caution}
:class: warning

📣 We have a new landing page!

We have a new landing page at <https://hubverse.io/> that provides a general overview of The Hubverse. Over the next few weeks, we will be removing duplicated
content from this site. 
```

````

File: docs/source/developer/index.md [hunk 2]
```diff
@@ -53,7 +53,7 @@ A summary of the approach and tools can be found in [The Hub Dashboard project p
 
 Discussions of new features and prioritizing bug fixes happens every Wednesday
 (except first Wednesday of the month) at the hubverse developer's meeting. Please
-contact Zhian Kamvar (zkamvar@umass.edu) if you would like to be invited.
+contact Anna Krystalli (r.rse.eu@gmail.com) if you would like to be invited.
```
zkamvar, line 56: Good catch!

annakrystalli:
Sure. I'm back on the project in Tuesday. Will have a closer look then 👍

zkamvar (commented):
> So I guess my overall comment is to be careful with moving and especially removing the docs too much before:
>
> * Checking the downstream implications
> * Making final and clear decisions about exactly what will live where

I agree with @annakrystalli on this. Renaming pages is effectively the same as deleting them without the benefit of reduced maintenance. We could set up redirects, but that's problematic as you would end up with redirects of redirects. 

I think the best thing to do right now is to put up a notice at the top of the index page that lets people know that the landing page is available.

## Background 

https://hubverse.io was motivated by https://github.com/hubverse-org/hubDocs/issues/211 and largely intended to replace the majority of the pages in the overview section (e.g. https://docs.hubverse.io/en/latest/overview/who-we-are.html is now https://hubverse.io/about.html). With that in mind, it does not make sense to keep two versions of the same information alive in two places.

The rationale behind the restructuring is because presenting things in long lists of information is not great for discoverability and understanding (the AWS console is a _really_ good example of what not to do). The two sites separate the concerns of "I want to know what the hubverse is" (hubverse.io) with "I want to know how to work with specific pieces of the hubverse" (docs.hubverse.io).

annakrystalli:
Thanks for the extra context @zkamvar . Honestly I think it's mainly the community and contributing guides that are linked to (and in time we could absolutely switch them over) but I wasn't 100% sure off the top of my head which is why I suggested checking. Overall it makes complete sense to move the overview section to the new site.

micokoch:
I will close this PR in favor of smaller PRs with fewer changes (and taking into consideration the input given above).

PR #342, opened 2025-05-21
add tip for downloading data from remote sources

zkamvar:
This is a part that would really help with testing changes:
understanding that you can fetch the data from the remote sources to
test instead of always starting from scratch.

This mostly affects how to test predtimechart changes because that's the
only tool that will avoid rebuilding existing data, but it's definitely
a useful skill

preview: https://hubdocs--342.org.readthedocs.build/en/342/developer/dashboard-local.html

### DIFF
--- a/docs/source/developer/dashboard-local.md
+++ b/docs/source/developer/dashboard-local.md
@@ -234,7 +234,7 @@ it should be feasible to fetch the hub as a sparse clone (see the tip in
 :class: note
 
 In this exercise, we are building a dashboard from scratch. However, live dashboards
-only ever start from scratch once in their lifetimes. For the rest of their life,
+only ever start from scratch once in their lifetimes. For the rest of their lives,
 they are working with _existing data_. If you want to mimic the remote dashboard
 process, **you can fetch these data from [their remote
 sources](#dashboard-local-remote-data) first**.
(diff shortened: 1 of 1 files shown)


docs/source/developer/dashboard-local.md
```diff
@@ -217,6 +229,56 @@ it should be feasible to fetch the hub as a sparse clone (see the tip in
 
 :::
 
+
+:::{admonition} Fetching data from the remote branches
+:class: note
+
+In this exercise, we are building a dashboard from scratch. However, live dashboards
+only ever start from scratch once in their lifetimes. For the rest of their life,
```
matthewcornell, line 237: Last word should be plural?: "For the rest of their lives, ..."

matthewcornell (approved):

docs/source/developer/dashboard-local.md
zkamvar:
```suggestion
only ever start from scratch once in their lifetimes. For the rest of their lives,
```

zkamvar: Please use the suggest icon to suggest minor changes like this in the future.

a red arrow pointing to the suggest icon in the WISYWIG menu

lshandross (approved):

PR #345, opened 2025-05-27
Minor updates to the checklist

micokoch:
Becky PR that was never merged (and seems reasonable to merge):
Add a note about forking the repo and be explicit about including tests.


LucieContamin:
Thanks! It looks good to me

PR #349, opened 2025-06-11
update information for new 5.1.0 schema

LucieContamin:
- update information for new 5.1.0 schema + add small description of the update

Link to [#130](https://github.com/hubverse-org/schemas/pull/130)


micokoch (approved):

docs/source/conf.py
```diff
@@ -87,10 +87,10 @@
 # -- Options for EPUB output
 epub_show_urls = "footnote"
 
-schema_version = "v5.0.0"
+schema_version = "v5.1.0"
 # Use schema_branch variable to specify a branch in the schemas repository from which config schema will be source, especially for docson widgets.
 # Useful if the schema being documented hasn't been released to the `main` branch in the schemas repo yet. If version has been released already, set this to "main".
-schema_branch = "br-" + schema_version
+schema_branch = "main" #"br-" + schema_version
```
annakrystalli, line 93: This should not be changed as when it's accompanied by a release the docs always build from the main branch. This is useful for previewing under development versions

annakrystalli (changes requested):

docs/source/conf.py
LucieContamin: We might want to also remove the "If version has been released already, set this to "main"." then!

annakrystalli: Agreed!

PR #370, opened 2025-08-15
Fix and update definitions to terminology.

micokoch:


LucieContamin (approved):
Thank you! It looks good!

PR #386, opened 2025-10-01
Update target date specs to match 2025-06-17-RFC-target-data-metadata decision

### LINKED ISSUE #382: Update target data documentation
We need to update our target data documentation to reflect the decisions made in https://github.com/reichlab/decisions/pull/31 

Specifically we need to:
- [ ] Clearly define observable unit and ensure this term is used consistently throughout
- [x] Add that an `as_of` column is now accepted in `oracle-output` data do but in limited to a single version of each unique combination of observable unit variable values
- [x] Remove any mention of a `date` column in time-series data which differs from the name of the column in the relevant column in the `oracle-output` data. Instead specify that such columns should share the same name in both target types.

annakrystalli:
A number of decisions were made in [2025-06-17-RFC-target-data-metadata](https://github.com/hubverse-org/hubDocs/issues/382) which affect expectations of target data and are not currently reflected in our current target data documentation.

This PR partially resolves #382 by:
- [x] Adding a note about an `as_of` column being allowed in oracle output data
- [x] Removes any mention of a `date` column in time-series and clarifies that columns indicating the date of observation must be consistent across `time-series` `oracle-output` and `model-output` data.

Regarding clear definition of the observable unit, I think the docs does a pretty good job already. It does however call it a unit of observation and includes the `observation` column in it while we have been limiting it to task IDs in the discussions in the decision file.  Some thoughts on how to address this discrepancy would be appreciated, especially by @nickreich but this can be done in #382 and addressed in a separate PR.

### DIFF
--- a/docs/source/user-guide/target-data.md
+++ b/docs/source/user-guide/target-data.md
@@ -85,21 +85,21 @@ target-data/
 The first format is *time series* data. This is often the native or
 "raw" format for data. Each row of the data set is a **unit of observation**, and the columns consist of:
 
-1. task ID variables that uniquely define the unit of observation. This must include at least one column representing the date.
+1. task ID variables that uniquely define the unit of observation. This must include at least one column representing the date of oservation. The column should share the same name across target data and model outputs.
 2. an `observation` column that records the observed value
 
 Here is an example of this form of data, showing selected dates for
 Massachusetts (FIPS code 25), drawn from the forecasting example in
 `hubExamples`:
 
-| date       | location | observation |
+| target_end_date       | location | observation |
 |:-----------|:---------|------------:|
 | 2022-11-19 | 25       |          79 |
 | 2022-11-26 | 25       |         221 |
 | 2022-12-03 | 25       |         446 |
 | 2022-12-10 | 25       |         578 |
 
-Here, the unit of observation is a date and location pair. That is, for each date and location, there is a single observed value.
+Here, the unit of observation is a target_end_date and location pair. That is, for each target_end_date and location, there is a single observed value.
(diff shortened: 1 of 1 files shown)


nickreich (approved):
The changes look good to me.

docs/source/user-guide/target-data.md
```diff
@@ -85,21 +85,21 @@ target-data/
 The first format is *time series* data. This is often the native or
 "raw" format for data. Each row of the data set is a **unit of observation**, and the columns consist of:
 
-1. task ID variables that uniquely define the unit of observation. This must include at least one column representing the date.
+1. task ID variables that uniquely define the unit of observation. This must include at least one column representing the date of oservation. The column should share the same name across target data and model outputs.
```
nickreich, line 88:
```suggestion
1. task ID variables that uniquely define the unit of observation. This must include at least one column representing the date of observation. The column should share the same name across target data and model outputs.
```

annakrystalli: Good catch! Fixed in d7d06b4a1e79c9858abc3e01c02f24e5a49662bb. Will open a separate PR

docs/source/user-guide/target-data.md
```diff
@@ -85,21 +85,21 @@ target-data/
...
 2. an `observation` column that records the observed value
```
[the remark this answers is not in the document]
nickreich: How about this change to the sentence above?
"Each row of the data set ~is~ corresponds to a **unit of observation**, where the columns consist of:"

I do think that a row corresponds to a unit of observation. And that unit is defined by the task-id variables. And then that unit has a value associated with it in the observation column.

PR #439, opened 2026-01-28
Improve sample output type page

### LINKED ISSUE #169: `sample-output-type.md`: Clarifying questions and editing suggestions
I am having a hard time understanding the sample output type chapter. One of the disconnects is that there are concepts that are presented before they are ever introduced (e.g. compound tasks and the implicit `compund_idx` column) and there are sections that are only pertinent to compound modelling tasks that are not subsections of that section.

## Introduction 

The introduction for the sample output type needs reworking. From what I've found in the historical documents, it seems that the text in the introduction was written before the sample schema was fleshed out:

https://github.com/hubverse-org/hubDocs/blob/a2738159c43b5e4051e9e0cebf87b293895447ac/docs/source/user-guide/sample-output-type.md?plain=1#L3-L23

When I read it, I wonder, "Why are we talking about the mean output type? This is the sample output type."

## Individual modeling tasks

Why is the `compound_idx` column here? It appears to be reiterating the grouping of the target column. Is this a column I should be worried about? The text says that it is implicit, but why does it have a name that indicates that it is a column that actually exists? 

Why is `column_idx` not defined in the schema?

## Compound modeling tasks

This description could be more specific to the example data presented, highlighting the columns in the text. 

> we will look at a hub reporting on variant proportions observed at a given location and at a given time. In the table below, a single modeling task is a unique combination of values from the task-id variables origin_date, horizon, variant, and location. In the table below, one set of four rows with the same values in the origin_date, horizon, and location columns but different variant values below represent four predicted variant proportions.

maybe:

> we will look at a hub reporting on **variant** proportions observed in **Massachusetts (`location`)** on **2024-03-15 (`origin_date`)** for **7 and 14 day forecasts (`horizon`)**. This is represented in the table below showing four variants (`AA`, `BB`, `CC`, and `DD`) represented over two horizons, giving us **eight unique modelling tasks.** 

What does "Base data: mean `output_type`" mean?

https://github.com/hubverse-org/hubDocs/blob/a2738159c43b5e4051e9e0cebf87b293895447ac/docs/source/user-guide/sample-output-type.md?plain=1#L70

### Four submissions

NOTE: For each submission, use level 4 headers, not bold text. 

Pain points: 

1. the schema says minimum of 90 samples, but we present two samples each. 
2. the sample numbers continue to increase across (as opposed to strictly within) tasks. Why?

#### Submission A

I am confused as to why the sample numbers keep increasing across the stratification and why there are only two samples per stratum. Should each stratum contain at least 90 samples (according to the schema).

#### Submission B

I think I understand now that this is showing two samples per stratum, but I'm still confused as to why the sample numbers continue to increase after change in strata. 

Are the values for each sample identical? 

#### Submission C

>  In this example, there is a single compound modeling task which we can describe as “Massachusetts with the origin_date of 2024-03-15”. 

"single compound modeling task" is confusing because there are two columns selected here. They do not vary, so it makes some sense in retrospect, but the initial read of this gives some roadblocks.

#### Submission D

The phrase "plain language" is a bit demotivating for a sentence with six prepositions. 

## Configuration of `output_type_id`

This description is clear (but it could use some trimming to reduce complexity) and the table illustrates the validity question better, but I feel that two things need to happen:

1. The section should be renamed to "Configuration of `compound_taskid_set`
2. The section should be a subsection of "Compound modelling tasks/"

I feel like this sentence is at conflict with the meaning of the schema (emphasis mine)

> A hub can specify a "compound_taskid_set" field in the metadata for the sample output_type to specify the task-id **columns that must be used to define** separate sample index values (as present in the output_type_id column). The following table shows how different specifications of this field would impact the validity of each of the example submissions A, B, C, and D. 

Given that submission C is valid for all of the schema configurations means that we should use "may" and not "must".

"columns that **may** be used to define"

## Number of samples

As I indicated above, it doesn't make sense why each task gets two samples. Also, I believe this belongs in the "Compound Modelling Tasks" section. 

## Relationship to output_types

I think this needs to be a subsection of "Compound Modelling Tasks".

lshandross:
Major overhaul of the sample output type docs page.

This doesn't require any code review, but I'd like a fairly detailed review to ensure the changes I've made are 1) accurate, 2) address the questions and comments in #169, and 3) understandable to the wider hubverse community (not just us developers). Likely it will be useful to have multiple people provide their reviews. Not time sensitive, but would be good to get merged in sooner rather than later


docs/source/user-guide/sample-output-type.md
```diff
@@ -1,42 +1,49 @@
...
+In the above table, the three task-id columns `origin_date`, `horizon`, and `location` uniquely define a modeling task in every row. Here, there are three modeling tasks (one for each row), represented by the tuples <br>
 
 ```
 {origin_date: "2024-03-15", horizon: "-1", location: "MA"}
 {origin_date: "2024-03-15", horizon: "0", location: "MA"}
 {origin_date: "2024-03-15", horizon: "1", location: "MA"}
 ```
 
-In words, the first of these tuples represents a forecast for one day (assume here the horizon is on the timescale of day) before the origin date of 2024-03-15 in Massachusetts.
+In words, the first of these tuples (as well as the first row in the table above it) represents a forecast for one day (assume here the horizon is on the timescale of day) before the origin date of 2024-03-15 in Massachusetts.
```
micokoch: Again, maybe you can say the first of these rows or lines.

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,42 +1,49 @@
...
-| ${\color{lightgrey}3}$ | 2024-03-15 | 1 | MA | sample | 8| - |
+| 1 | 2024-03-15 | -1 | MA | sample | 0| - |
+| 1 | 2024-03-15 | -1 | MA | sample | 1| - |
+| 1 | 2024-03-15 | -1 | MA | sample | 2| - |
+| 2 | 2024-03-15 | 0 | MA | sample | 3| - |
+| 2 | 2024-03-15 | 0 | MA | sample | 4| - |
+| 2 | 2024-03-15 | 0 | MA | sample | 5| - |
+| 3 | 2024-03-15 | 1 | MA | sample | 6| - |
+| 3 | 2024-03-15 | 1 | MA | sample | 7| - |
+| 3 | 2024-03-15 | 1 | MA | sample | 8| - |
```
micokoch: Across all the tables, I would prefer to have some values included for the mean, to better illustrate the point. There is a date chosen, a horizon, and a location - couldn't example values be given for the mean, which could be compared with the first table to illustrate the point.

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,86 +64,104 @@ In this setting, a hub will specify a minimum and maximum number of required sam
 }
 ```
 
-In words, the above configuration specifies that  `"output_type_id_params"` samples are required, they must be integers, and there must be exactly (i.e., no more or less than) 100 samples per modeling task. The "value" specifications correspond to the values contained in the "value" column (e..g they must be storable as numeric "double" format and be no less than zero).
+In words, the above configuration specifies that  `"output_type_id_params"` samples are required, they must be integers, and there must be exactly (i.e., no more or less than) 3 samples per modeling task. The "value" specifications correspond to the values contained in the "value" column (e.g., they must be storable as numeric "double" format and be no less than zero).
 
 Note that the `output_type_id` parameters are specified in an `"output_type_id_params"` block because they are parameters defining the allowable values. For other output types, the `"output_type_id"` block is used to list required and optional values explicitly.
```
micokoch: Personally, I would like a small chunk of sample code in the example above that shows what is contained in `"output_type_id"` and where it falls in the schema hierarchy. So, for instance show the `"properties"` of the mean.

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,86 +64,104 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
+
+[^2]: In the tabular data, an entry of "-" stands in for specific values to be provided by the submitter. They are not assumed to be identical.
 
-### Four submissions, differing by compound modeling task
-**Submission A**: sample `output_type` where **a single modeling task corresponds to a unique combination of `origin_date`, `location`, `horizon`, and `variant`**. There are eight unique modeling tasks in this example.
+Notice that this example submission file could be displaying predictions with any number of dependence structures (or none at all) since we are not given the `output_type_id` values or compound task-id set. The following subsection provides four possible combinations of compound modeling tasks that this tabular data may be representing.
 
+### Examples of distinct compound modeling tasks
+
+#### Example A: No task ids display dependence
```
micokoch: Consistency - Task IDs

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,86 +64,104 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
+    "max_samples_per_task": 2,
     "compound_taskid_set": ["origin_date", "location", "horizon", "variant"]
 }
 ```
 
+The table below displays a subset of a possible submission for the prediction task specified in the above schema. There are eight unique modeling tasks in this example, with two samples for each.
+
 ```{attention}
-Rows are shaded to indicate different samples for the same compound forecast task. For example, in the table below, the first two rows correspond to the same modeling task, but they are highlighted to show that the first row is from sample "s0" while the second row is from sample "s1".
+Rows with the same `compound_idx` value indicate distinct sample forecasts made for the same compound forecast task. For example, in the table below, each pair of rows with the same `compound_idx` correspond to the same modeling task, but are each from one of sixteen distinct sample draws "s#".
```
micokoch: - I find this phrase "one of sixteen distinct sample draws "s#"" confusing. I would phrase it as "but are each from distinct sample draws (e.g., `s0` and `s1` are two distinct sample draws for {origin_date: "2024-03-15", horizon: "7", location: "MA", variant: "AA"}).
- I am a little thrown off by the Python numbering, but if it makes sense for most modelers, that's fine.

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,86 +64,104 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
 | 2024-03-15 | 14 | DD | MA | sample | - | - |
+| 2024-03-15 | 14 | DD | MA | sample | - | - |
+
+[^2]: In the tabular data, an entry of "-" stands in for specific values to be provided by the submitter. They are not assumed to be identical.
 
-### Four submissions, differing by compound modeling task
-**Submission A**: sample `output_type` where **a single modeling task corresponds to a unique combination of `origin_date`, `location`, `horizon`, and `variant`**. There are eight unique modeling tasks in this example.
+Notice that this example submission file could be displaying predictions with any number of dependence structures (or none at all) since we are not given the `output_type_id` values or compound task-id set. The following subsection provides four possible combinations of compound modeling tasks that this tabular data may be representing.
 
+### Examples of distinct compound modeling tasks
```
micokoch: In general, these examples are very clear, but one general comment is that at the beginning I would say something like "different samples are distinguished from each other by using light grey and grey to highlight adjacent samples." This holds true for all examples, except Example A, where you would have to highlight all odd samples in grey (perhaps changing the css heatMap1 configuration).

docs/source/user-guide/sample-output-type.md
```diff
@@ -293,30 +325,33 @@ https://www.w3.org/WAI/tutorials/tables/irregular/#table-with-two-tier-headers
...
+These derived task-ids must be properly configured, or they can cause problems when validating compound modeling tasks by throwing erroneous errors. *If **all** the task-id variables a derived task-id is derived from are part of the `compound_task_id_set`, then that derived task-id must also be a part of the `compound_task_id_set`; otherwise, that derived task-id should be excluded.*
 ```
 
-## Number of samples vs. `output_type_id`
+### Number of samples vs. `output_type_id`
+
 The number of samples per individual modeling task in the above examples can always be determined by the number of times that each unique combination of task-id variables (i.e., each individual modeling task) appears in the submission. For Submissions A, B, C, and D above, even though the number of unique values of `output_type_id` changes, all examples have two samples per individual modeling task since each task-id-set appears exactly twice in the provided data.
```
micokoch: I would add a parenthesis at the very end : "(i.e., a unique combination of "origin_date", "location", "horizon", "variant").

micokoch (approved):
Overall this is a good improvement, and you address all the comments Zhian had made. It's still a bit tricky to follow, but that may be the nature of the beast. I made various comments and suggestions, and I hope they make sense. My overall suggestions are:
- Ensure consistency with terms. Task ID is my main issue, as it is written differently in different parts, and I wish we were consistent across hubDocs.
- I think you should have a mini-glossary or a summary at the very beginning of key terms like `compound_idx`, `task ID`, etc. I know this would be duplicative, but as I was reading, I thought it would be helpful to go back to the beginning and see what those concepts mean.
- Finally, I really think you should include values in the `value` column. It would help illustrate.
Great job, and I hope others comment, as it's important to get many eyes to ensure these explanations are clear.

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,86 +64,104 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
 
 (compound-modeling-tasks)=
 ## Compound modeling tasks
 
-In some settings, modeling hubs may wish to identify sets of modeling tasks that the hub will treat as related, for example, when multiple distinct values can be seen as representations of a single multivariate outcome of interest. In these settings, a subset of the task-id columns (a `"compound_taskid_set"`) will be used to identify what values are shared for the modeling tasks related to each other.
+In some settings, modeling hubs may wish to identify sets of modeling tasks that the hub will treat as related; for example, when multiple distinct values can be seen as representations of a single multivariate outcome of interest (i.e., forecasts originate from a joint predictive distribution). We use a subset of the task-id columns (the `"compound_taskid_set"`) to identify which variables do not display any dependence within or across their values. Any samples with the same output_type_id are assumed to to be a single sample from a joint distribution.
```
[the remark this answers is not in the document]
lshandross: What do you think of the changes Anna made to the language:

> In both cases, joint sampling introduces additional dimensions to the predictive distribution. Instead of a univariate distribution at each modeling task, we have a multivariate distribution spanning multiple modeling tasks. We refer to this as **response dependence** across the task IDs that vary within a group, because the predicted values (responses) for different modeling tasks are statistically dependent.
> 
> Within a group of jointly sampled modeling tasks, some task ID values remain constant (defining which group we're in), while others vary (defining the multiple modeling tasks covered by the joint distribution). The `"compound_taskid_set"` specifies which task IDs remain constant within a group. Task IDs not in this set vary within the group and are sampled jointly, introducing response dependence across those task IDs. A **compound modeling task** is thus a group of related predictions defined by the compound task ID set.

docs/source/user-guide/sample-output-type.md
```diff
@@ -145,107 +170,114 @@ Once again, rows are grouped so each unique sample for each modeling task is tog
...
-    "max_samples_per_task": 100,
+    "min_samples_per_task": 2,
+    "max_samples_per_task": 2,
     "compound_taskid_set": ["origin_date", "location"]
 }
 ```
 
+The table below has one unique compound modeling task (shown with the `compound_idx` column value) and two unique sample draws. Each sample represents a grouped collection of possible values for all four variants across both prediction horizons.
+
 <div class="heatMap3">
```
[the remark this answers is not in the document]
lshandross: hmm, this also isn't working. Any other suggestions?

docs/source/user-guide/sample-output-type.md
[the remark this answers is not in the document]
micokoch: This is complicated, and so it will probably always be a bit hard to follow. I had to read the different versions twice to make sure I understood, and in all cases, I got it. By now, I'm not sure what is the clearest formulation, but both Anna and your approaches seem good to me, so I think you should decide what reads best. Thanks for asking.

docs/source/user-guide/sample-output-type.md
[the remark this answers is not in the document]
micokoch: I'm looking at the preview now, and it looks correct (if you mean to highlight each sample, which is what I think you meant to do).

docs/source/user-guide/sample-output-type.md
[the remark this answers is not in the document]
lshandross: I went ahead and changed the numbering scheme

docs/source/user-guide/sample-output-type.md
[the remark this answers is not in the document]
lshandross: Thanks for your input! I've already accepted Anna's changes, so I'll go with her version

docs/source/user-guide/sample-output-type.md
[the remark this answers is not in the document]
lshandross: Ah, it looks like it just hadn't updated yet on my end. I can see the changes now!

lshandross:
@micokoch I'm still working on adding values for the `value` column, but I want to get your opinion on the section on key terms with definitions. Are there any terms that I should add? And what do you think of the definitions I provided? (I tweaked the definitions for the terms already given elsewhere in the docs to fit the information on this page)

micokoch:
@lshandross - thanks for the changes, Li! I'm way past my 5 weekly hours for this week, so I'll take a closer look at this next week. Hope that's okay, but let me know if you need it sooner.

lshandross:
@micokoch Some other questions about the values for the example model outputs: For the section with the four examples, should the sample prediction values be the same in all cases (i.e., the value column is identical across examples A-D)? Or would it be more useful for them to be different?

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
 
-In the above table, the three task-id columns `origin_date`, `horizon`, and `location` uniquely define a modeling task. Here, there are three modeling tasks, represented by the tuples <br>
+## Definitions of key terms
+
+[modeling-tasks]{#key-term-modeling-tasks}
+: A definition of the goals of a modeling effort (i.e., what it is hoping to predict), possibly including conditions, assumptions, and task ID variables. Generally, they are defined by unique combinations of the task IDs.
+
+[task IDs (AKA task ID variables, task ID columns)]{#key-term-task-ids}
+: Column(s) in model output that provide details about what is being predicted. These "task ID" columns may also include additional information, such as any conditions or assumptions used to generate the predictions. Common examples include `target`, `location`, `reference_date`, and `horizon`.
```
nickreich: I'd suggest moving the middle sentence about additional information to the end of this paragraph and maybe adding an additional example here, as I think this is taken from a scenario hub?

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+
+[output type ID]{#key-term-output-type-id}
+: A column in model output that specifies more identifying information specific to the output type. For samples, an integer or string providing a unique identifier for the sample. If one or more task IDs display of response dependence, rows of sample predictions from a particular model that share an output type ID may be assumed to represent a single sample from a joint distribution across multiple levels of the task ID variables.
```
nickreich: I don't understand the sentence that starts "If one or more ..."

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+
+[compound_idx]{#key-term-compound_idx}
+: A column used a visual aid to indicate which rows belong to the same group in the example model output on this page. In the marginal case, each group contains samples for one modeling task. The `compound_idx` column is not a task ID variable and is not typically present in actual model output data.
```
nickreich: it's not just a "visual aid" right, it's actually like a specific identifer in this example? This is not clear from this point in the description.

nickreich: it sort of feels like these definitions are coming sort of early in the description?

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+## Sampling modeling tasks
 
-## Individual modeling tasks
-In many settings, forecasts will be made for individual modeling tasks, with no notion of modeling tasks being related to each other or collected into sets (for more on this, see the [compound modeling tasks section](#compound-modeling-tasks)). In the situations where forecasts are assumed to be made for individual modeling tasks, every modeling task is treated as distinct, as is implied by the `compound_idx` column in the table below (grayed out to indicate that such a column exists implicitly in the dataset and is not typically present in the actual tabular data). In this setting, the `output_type_id` column indexes the samples that exist for each modeling task.
+While the mean output type produces a single predicted value per modeling task, the sample output type captures uncertainty by providing multiple possible values at each slice of task ID space. Each sample represents one possible outcome, and together, the collection of samples describes a distribution of predicted values.
```
nickreich: I don't love the "slice of task ID space" language. It feels overly jargony.
```suggestion
While the mean output type produces a single predicted value per modeling task, the sample output type captures uncertainty by providing multiple possible values for each task. Each sample represents one possible outcome, and together, the collection of samples describes a distribution of predicted values.
```

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+
+How modeling tasks relate to each other determines the structure of this distribution. When modeling tasks are treated independently, samples at each slice form a univariate (single-variable) distribution. When modeling tasks are grouped together, samples capture a multivariate (joint) distribution across the group.
```
nickreich:
```suggestion
How modeling tasks relate to each other determines the structure of this distribution. When modeling tasks are treated independently, samples for each task form a univariate (single-variable) distribution. When modeling tasks are grouped together, samples capture a multivariate (joint) distribution across the group of tasks.
```

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+
+## Marginal distributions
+
+In many settings, predictions will be made for individual modeling tasks, with no notion of modeling tasks being related to each other or collected into sets (for more on this, see the [compound modeling tasks section](#compound-modeling-tasks)). When predictions are assumed to be made for individual modeling tasks, every modeling task is treated as distinct. In mathematical terms, these samples represent draws from marginal predictive distributions.
+
+Now, suppose we wanted to collect samples for each of the modeling tasks defined in the previous section. The table below shows a dataset with three groups (indicated by `compound_idx`[^1] values 1, 2, and 3) and three samples per group. The `output_type_id` column contains sample indexes that are unique across an entire model output file, not just within each group, so each group has distinct sample indexes (1-3, 4-6, and 7-9 respectively). When sampling from marginal distributions, each group corresponds to a single modeling task, so all task ID values are identical within each group. Notice how each group represents samples for a single slice of task ID space.
```
nickreich:
```suggestion
Now, suppose we wanted to collect samples for each of the modeling tasks defined in the previous section. The table below shows a dataset with three groups (indicated by `compound_idx`[^1] values 1, 2, and 3) and three samples per group. In this case, the only difference between the modeling tasks here are the horizon. The `output_type_id` column contains sample indexes that are unique across an entire model output file, not just within each group, so each group has distinct sample indexes (1-3, 4-6, and 7-9 respectively). When sampling from marginal distributions, each group corresponds to a single modeling task, so all task ID values are identical within each group. Notice how each group represents samples for a single modeling task.
```

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+| 1 | 2024-03-15 | 1 | MA | sample | 2| - |
+| 1 | 2024-03-15 | 1 | MA | sample | 3| - |
+| 2 | 2024-03-15 | 2 | MA | sample | 4| - |
+| 2 | 2024-03-15 | 2 | MA | sample | 5| - |
+| 2 | 2024-03-15 | 2 | MA | sample | 6| - |
+| 3 | 2024-03-15 | 3 | MA | sample | 7| - |
+| 3 | 2024-03-15 | 3 | MA | sample | 8| - |
+| 3 | 2024-03-15 | 3 | MA | sample | 9| - |
+
+In this setting, a hub will specify a minimum and maximum number of required samples per group in the configuration for the prediction task. In the marginal case, each group corresponds to a single modeling task, but as we will see in the [compound modeling tasks section](#compound-modeling-tasks), a group can span multiple modeling tasks. The associated configuration might look like:
```
nickreich:
```suggestion
In this setting, a hub will specify a minimum and maximum number of required samples per group in the configuration for the prediction task. In this "marginal" case, each group corresponds to a single modeling task, but as we will see in the [compound modeling tasks section](#compound-modeling-tasks), a group can span multiple modeling tasks. The associated configuration might look like:
```

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,195 +82,254 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
+        },
+        "is_required": true
+    }
+```
 
 (compound-modeling-tasks)=
 ## Compound modeling tasks
 
-In some settings, modeling hubs may wish to identify sets of modeling tasks that the hub will treat as related, for example, when multiple distinct values can be seen as representations of a single multivariate outcome of interest. In these settings, a subset of the task-id columns (a `"compound_taskid_set"`) will be used to identify what values are shared for the modeling tasks related to each other.
+In the previous section, we saw that when sampling from marginal distributions, each sample is drawn from a single modeling task (a single slice of task ID space). In some settings, however, modeling hubs may wish to capture relationships between modeling tasks by sampling from a joint distribution. This means drawing values for multiple modeling tasks at once as a coherent set.
```
nickreich:
```suggestion
In the previous section, we saw that when sampling from marginal distributions, each sample is drawn from a single modeling task. In some settings, however, modeling hubs may wish to capture relationships between modeling tasks by sampling from a joint distribution. This means drawing values for multiple modeling tasks at once as a coherent set.
```
If other people find the "slice" language to be helpful then I will defer. I mostly find it distracting when we have this other clear concept of a "modeling task" in place already. I think we should just use the clear term since we have one for it.

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,195 +82,254 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
+- **Variant proportions**: A model predicting proportions of multiple disease variants might draw all variant proportions together, a sample from a joint distribution across variants.
 
-As a running example of how compound modeling tasks could be specified differently, we will look at a hub reporting on variant proportions observed at a given location and time. In the table below, a single modeling task is a unique combination of values from the task-id variables `origin_date`, `horizon`, `variant`, and `location`.  In the table below, one set of four rows with the same values in the `origin_date`, `horizon`, and `location` columns, but different variant values below represent four predicted variant proportions.
+In both cases, joint sampling introduces additional dimensions to the predictive distribution. Instead of a univariate distribution at each modeling task, we have a multivariate distribution spanning multiple modeling tasks. We refer to this as **response dependence** across the task IDs that vary within a group, because the predicted values (responses) for different modeling tasks are statistically dependent.
```
nickreich:
```suggestion
In both cases, joint sampling introduces additional dimensions to the predictive distribution. Instead of a univariate distribution at each modeling task, we have a multivariate distribution spanning multiple modeling tasks. We refer to this as **dependence** across the task IDs that vary within a group, because the predicted values for different modeling tasks are statistically dependent.
```

docs/source/user-guide/sample-output-type.md
```diff
@@ -57,195 +82,254 @@ In this setting, a hub will specify a minimum and maximum number of required sam
...
+In the previous section, we saw that when sampling from marginal distributions, each sample is drawn from a single modeling task (a single slice of task ID space). In some settings, however, modeling hubs may wish to capture relationships between modeling tasks by sampling from a joint distribution. This means drawing values for multiple modeling tasks at once as a coherent set.
+
+Consider two common scenarios:
+- **Trajectories over time**: A model might predict values across multiple time horizons as a coherent path. Rather than drawing each horizon's prediction independently, the model draws an entire trajectory, a sample from a joint distribution over horizons.
+- **Variant proportions**: A model predicting proportions of multiple disease variants might draw all variant proportions together, a sample from a joint distribution across variants.
```
nickreich, line 93: We could consider adding locations as another (I think more intuitive) dimension in addition to trajectories over time. E.g., does a model simulate trajectories across space and time together? or for each location separately?

docs/source/user-guide/sample-output-type.md
```diff
@@ -274,49 +358,54 @@ https://www.w3.org/WAI/tutorials/tables/irregular/#table-with-two-tier-headers
...
 
 ```{caution}
-#### `"compound_taskid_set"` and derived task-id variables
+**Derived task IDs** are a type of task IDs whose values depend wholly on that of other task ID variables. A common example is the `target_end_date` task ID, which tends to be derived from the combination of the `reference_date` (or `origin_date`) and `horizon` task IDs.
 
-There is a class of task-ids that can cause problems for validation of compound modeling tasks if not properly configured, that of **derived task-ids** i.e. task-ids whose values depend on the values of other task-id variables. An example is the `target_end_date` task-id which is most commonly derived from the combination of the `reference_date` or `origin_date` and `horizon` task-ids.
+These derived task IDs must be properly configured, or they can cause problems when validating compound modeling tasks by throwing erroneous errors. *If **all** the task ID variables a derived task ID is derived from are part of the `compound_taskid_set`, then that derived task ID must also be a part of the `compound_taskid_set`; otherwise, that derived task ID should be excluded.*
```
nickreich:
```suggestion
These derived task IDs must be properly configured, or they can cause problems when validating compound modeling tasks by throwing errors when they should not. *If **all** the task ID variables a derived task ID is derived from are part of the `compound_taskid_set`, then that derived task ID must also be a part of the `compound_taskid_set`; otherwise, that derived task ID should be excluded.*
```

nickreich (commented):
I have a handful of mostly minor phrasing comments.

docs/source/user-guide/sample-output-type.md
lshandross: 1. What do you mean by specific identifier? My understanding is that the compound index column is not a task ID, although I was not one of the people who created the example data

2. This section is meant as a sort of glossary-like section that people can refer back to. Some of the terms are defined later in the page, while others are from earlier pages in the documentation. I could move this section to the end but note its existence in one of the first few sections instead

docs/source/user-guide/sample-output-type.md
lshandross: I don't have much of an opinion on this particular wording, as Anna was the one who introduced it in her changes. I'm happy to defer to majority opinion from the others who have reviewed or edited this PR

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
 | 2024-03-15 |  1 | MA | mean | NA| - |
+| 2024-03-15 |  2 | MA | mean | NA| - |
+| 2024-03-15 |  3 | MA | mean | NA| - |
 
+In this table, the task ID columns are `origin_date`, `horizon`, and `location`. There are three modeling tasks (one per row): horizon 1, horizon 2, and horizon 3—all for Massachusetts and origin date 2024-03-15. The first row, for example, represents a single slice of task ID space, a prediction for one week ahead in Massachusetts, resulting in one predicted value.
 
-In the above table, the three task-id columns `origin_date`, `horizon`, and `location` uniquely define a modeling task. Here, there are three modeling tasks, represented by the tuples <br>
+## Definitions of key terms
+
+[modeling-tasks]{#key-term-modeling-tasks}
```
micokoch: I don't think modeling tasks should have a hyphen. It isn't used in the modeling tasks section.

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
-| 2024-03-15 |  0 | MA | mean | NA| - |
 | 2024-03-15 |  1 | MA | mean | NA| - |
+| 2024-03-15 |  2 | MA | mean | NA| - |
+| 2024-03-15 |  3 | MA | mean | NA| - |
 
+In this table, the task ID columns are `origin_date`, `horizon`, and `location`. There are three modeling tasks (one per row): horizon 1, horizon 2, and horizon 3—all for Massachusetts and origin date 2024-03-15. The first row, for example, represents a single slice of task ID space, a prediction for one week ahead in Massachusetts, resulting in one predicted value.
 
-In the above table, the three task-id columns `origin_date`, `horizon`, and `location` uniquely define a modeling task. Here, there are three modeling tasks, represented by the tuples <br>
+## Definitions of key terms
+
```
micokoch: Thanks for writing all these definitions, Li! I notice you use those curly bracket anchor links (e.g., `{#key-term-modeling-tasks}`). They're not consistent with the format I have used (`(key-term-modeling-tasks)=`), but I see they've been used in the developer guide and the dashboard documentation, so it seems fine. More importantly, I think they're unnecessary, unless you plan on linking to them at a later date. I would wait until then to create them, but that's just a suggestion.

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+[modeling-tasks]{#key-term-modeling-tasks}
+: A definition of the goals of a modeling effort (i.e., what it is hoping to predict), possibly including conditions, assumptions, and task ID variables. Generally, they are defined by unique combinations of the task IDs.
```
micokoch: It's great that you matched the definitions to what's in the terminology page!

docs/source/user-guide/sample-output-type.md
```diff
@@ -1,52 +1,77 @@
...
+
+[task IDs (AKA task ID variables, task ID columns)]{#key-term-task-ids}
+: Column(s) in model output that provide details about what is being predicted. These "task ID" columns may also include additional information, such as any conditions or assumptions used to generate the predictions. Common examples include `target`, `location`, `reference_date`, and `horizon`.
+
+[output type ID]{#key-term-output-type-id}
```
micokoch: Here I think it would help to use the underscores (as you use them later), so that it matches the column name (i.e., "output_type_id").

docs/source/user-guide/sample-output-type.md
micokoch: Here is my attempt:
```suggestion
: A column in model output that specifies identifying information specific to the output type. For samples, an integer or string providing a unique identifier for the sample. This is important for instances when samples come from a joint distribution, so that some rows may not share all task ID values, but if they have the same `output-type_id`, they are presumed to represent a single sample from a joint distribution. Examples are given below.
```

docs/source/user-guide/sample-output-type.md
micokoch: I will chime in, only because I think I suggested Li make a glossary. Thank you, I think it's really good, but maybe my idea wasn't so good. The concepts are so complicated and convoluted that perhaps reading these definitions early on may confuse more than clarify. A case in point for me is the relationship between `compound_idx` and `output_type_id`. I think I only understood this when I read the examples and saw that the you can have various `output_type_id`s for a single `compound_idx`, but not the other way around. Maybe the glossary is just not a good idea. I always have to read these sentences various times to understand, but maybe that's the nature of the beast.

micokoch: A separate comment - I find the use of "marginal" confusing at times, because it's used on its own (and the word has a couple of meanings). I would always say something like "in the case of a marginal distribution" or use some other qualifier (such as "sampled") to make it clear. This is what is done with "joint distributions", and it might help to contrast them early on?

micokoch (approved):
This is really good work @lshandross , and I'm just going to approve, because mine are only suggestions. I tried not to give too much feedback, because too many cooks might make this more confusing, but I want to address some of your prior comments:
1. I gave some thoughts on the key terms, and now I'm thinking maybe it wasn't a good idea (my bad). It's just so helpful to see the examples, that on their own, they may be more confusing.
2. As for keeping the value column the same or different, I think that it makes sense to keep this as real as possible, so if you had real values, that would be great. But if not, I would have them be similar, but not always identical, because that drives the point that the results from a joint distribution may vary from those of a marginal distribution, at least partly because of the dependence. I mean, if you see a case where it seems that keeping the values almost identical helps illustrate and explain better, then you should do that. But generally, my expectation would be that they would be different. I hope that helps.
3. My feeling on the slices is that it's probably easier to read fewer words, so I would go with Nick's edits and just remove the concept from the page.
Thanks for your work on all of this, and I would say that if you feel satisfied with the edits the next time, just merge it. Nothing is sticking out as being wrong, and there's always the possibility to make future edits. This is already a big improvement. Thanks!

docs/source/user-guide/sample-output-type.md
lshandross: I think that my final decision with this will be to remove the glossary, but perhaps create a footnote and/or link to a couple of terms that people might want refresher on that are not defined on the page earlier so people don't have to go hunting for term definitions elsewhere in the documentation

nickreich: My point about the specific identifier is that the column 'compound_idx' doesn't seem merely as a "visual" thing, but the numbers are actually indexing different compound task ids, right?

lshandross: That's true, but I guess I was trying to get across the fact that the `compound_idx` isn't 1) a task id in any of the examples, 2) typically included in model output as a (task id) column, 3) or really necessary to include in model output to delineate compound tasks once you understand the concept behind the compound task id set.

Would you prefer that I remove the term "visual aid" from the definition? I'm also open to other suggestions to make this clearer

nickreich:
```suggestion
: A column used to indicate which rows belong to the same group in the example model output on this page. In the marginal case, each group contains samples for one modeling task. The `compound_idx` column is not a task ID variable and is not typically present in actual model output data.

```

lshandross:
I ended up filing an issue for the idea of adding actual values to the example data on this page, instead of implementing it in this PR. I was getting bogged down by trying to make sure I chose reasonable values for all the data (since we don't have predictions for these particular model tasks that I can pull from), so I felt it was better to just get this merged in and worry about the values later

PR #474, opened 2026-05-29
Fixing failed Dependabot checks mentioned in #473

### LINKED ISSUE #473: Change Dependabot configuration
The GitHub Actions Dependabot updates are not working because this site is not built that way. Action items:
- Fix the actions that are regularly failing.
- Check the way the website is being built to review if other Dependabot-style updates are needed.
- Review how to trigger a website build.

micokoch:
Fix #473

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,15 +1,16 @@
-# instruct GitHub dependabot to scan github actions and Python code for dependency updates
+# instruct GitHub dependabot to scan pre-commit hooks and Python code for dependency updates
+# note: this site is built by Read the Docs (not GitHub Actions), so there are no workflow
+# files for the github-actions ecosystem to scan
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    # dependabot automatically checks .github/workflows/ and .github/actions/
+  - package-ecosystem: "pre-commit"
     directory: "/"
     schedule:
       interval: "weekly"
     # group all run-of-the mill updates into a single pull request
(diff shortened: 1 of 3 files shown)


micokoch:
This PR fixes the regularly failing `github_actions in /. - Update` Dependabot job. The failure was caused by the `github-actions` ecosystem being configured in `dependabot.yml` but having no `.github/workflows/` files to scan — so every run found nothing and failed.

The fix replaces the `github-actions` ecosystem with `pre-commit`, so Dependabot will now keep hook versions in `.pre-commit-config.yaml` up to date automatically. The pre-commit hook versions are also manually bumped to their current releases as a starting point.

annakrystalli (approved):
Good catch @micokoch ! Am approving. One question I have, if we do end up introducing workflows in the future we might want to revert or amend this. Wondering if you had any ideas on where we could maybe add a note documenting this and prompting to take appropriate action? Perhaps I am overthinking this though...😜

micokoch:
Hi @annakrystalli - thank you for reviewing (and approving) the PR. Just to be clear, it was my mistake to add that Dependabot recurring action, as I just copied what was happening in other repos. So, this was fixing my own mistake. I will write an issue on your comment, and just label it as "on hold," so that if we do shift to workflows, someone can see the issue.

annakrystalli:
Ah ok! Don't worry about it then. I think I'm overthinking it

micokoch:
Okay! Then I won't write an issue ;)

PR #477, opened 2026-06-02
Document scale transformations in the PredEvals user guide

### LINKED ISSUE #464: Document scale transformation configuration in the dashboards user guide
Part of tracking issue hubverse-org/hubPredEvalsData#34.

## Scope

Extend `docs/source/user-guide/dashboards.md` with a "Scale transformations" section covering:

- **Why** — evaluating count forecasts whose magnitudes span orders of magnitude.
- **`transform_defaults`** — top-level default applied to all targets.
- **Per-target `transform`** — override semantics (**replaces defaults entirely — no merge**).
- **Opt-out** with `transform: false`.
- **Supported function allowlist** — `log_shift` (scoringutils), `sqrt`, `log1p`, `log`, `log10`, `log2`. Include argument tables and links to R function docs. Base of logarithm stated explicitly for `log1p` to avoid ambiguity.
- **Output behaviour** — wide-format `scores.csv`, one row per model per disaggregation key, transformed scores as labelled columns (`log_wis`, `log_ae_median`, …). No `scale` column.
- **pmf / categorical output types** — transform silently skipped; behaviour differs between explicit (errors) vs inherited (warns) configuration.
- **How transform interacts with output types in the same target** — one transform per target applies uniformly to every transformable output type the target declares (quantile / mean / median today; sample once Sprint C lands). pmf output types in the same target are scored on the natural scale regardless, producing `NA` in `log_*` columns for those rows. Link to the hubEvals metric × output-type compatibility reference so admins can see which metrics apply where.
- **Metrics that don't change under transformation.** All v1.1.0 transforms are strictly monotone. A couple of default quantile metrics are *invariant* under such transforms — their `log_*` columns equal the natural counterpart column-for-column and carry no new information:
  - `interval_coverage_50`, `interval_coverage_90` — interval containment is preserved under monotone transforms.
  - `bias` — direction/rank-of-error is invariant.

  These duplicate columns are expected and not filtered out (matches upstream scoringutils/hubEvals behaviour). Users can safely ignore the `log_*` variants of these metrics. Metrics that DO produce genuinely different values under transform include `wis`, `overprediction`, `underprediction`, `dispersion`, `ae_median`, and point-error metrics.
- **Minimal worked example YAML.**

## Verification

Docs build locally without errors; cross-links to the hubPredEvalsData v1.1.0 schema and release notes resolve.

annakrystalli:
## Summary

Documents the **scale transformations** feature for the PredEvals evaluation module, extending the "Scale transformations" section in `docs/source/user-guide/dashboards.md`.

Closes #464.

> [!IMPORTANT]
> **The feature is not fully available yet.** `hubPredEvalsData` v1.1.0 (the release that introduces transform support) is still being finalised, and the dashboard-side presentation is in progress. This PR **queues the docs up in preparation** for the release and to **elicit early review**. It should land alongside (or just after) the v1.1.0 release rather than ahead of it.

## What's in here

The section was reconciled against `hubPredEvalsData` v1.1.0 as actually implemented (schema, scoring, options generation, and the still-open invariant-metric gating PR), and reframed around what hub admins see in the dashboard rather than backend file formats:

- **Supported transform functions** — full allowlist (`log_shift`, `sqrt`, `log1p`, `log`, `log10`, `log2`) with sources, R-doc links, and the `log_shift` offset note.
- **Which output types are transformed** — `mean`/`median`/`quantile` yes; `sample` under consideration; `pmf` not transformable.
- **How transformed scores appear in the dashboard** — the parenthetical label presentation (e.g. `WIS (log)` alongside `WIS`) and `append` behaviour, with the invariant metrics (interval coverage, bias) shown once.
- **Validation behaviour** — explicit transform on a non-transformable target errors; inherited `transform_defaults` warns and is skipped.

## Review asks

- Is the level of detail right for the user guide, or too much?
- Wording/framing on the dashboard presentation, especially before the predevals UI is finalised.

## Verification

- Docs build locally (`sphinx-build`) with no new warnings.
- `codespell` and `trailing-whitespace` pre-commit hooks pass on the changed file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/user-guide/dashboards.md
+++ b/docs/source/user-guide/dashboards.md
@@ -385,7 +385,7 @@ In principle, it would be possible to implement `round_filters` by specifying th
 :::
 
 (predevals-scale-transformations)=
-### Scale transformations (optional)
+### Scale transformations for PredEvals (optional)
 
 When a target's values span a wide range of magnitudes, scores computed on the natural scale can be dominated by the largest values. Applying a scale transformation before scoring can give a more balanced evaluation across magnitudes. PredEvals supports configuring scale transformations in `predevals-config.yml` from schema version `v1.1.0` onwards, applied instead of or alongside natural-scale scoring.
 
@@ -400,17 +400,18 @@ Transforms can be configured at two levels:
 - **`transform_defaults`** (top-level, optional): a default transformation applied to every target whose available output types support transformation.
 - A **`transform`** field on an individual target (per-target, optional): a transformation specific to that one target, which *replaces* `transform_defaults` entirely for it.
 
-For example, the abbreviated config below sets a default transform, has the first target inherit it, the second target use its own transform instead, and the third opt out:
+The structure of a `transform` block and the functions available are detailed below, but it may be easier to start with an illustrative case. In the abbreviated config below, a default transform is set, the first target inherits it, the second target uses its own transform instead, and the third opts out:
 
 ```yaml
 transform_defaults:          # applies to all transformable targets by default
   fun: log_shift
   args:
     offset: 1
+  label: log                 # dashboard label for the transformed scale, e.g. "WIS (log)"
 targets:
 - target_id: wk inc flu hosp
   metrics: [wis]
-  # no transform field -> inherits transform_defaults (log_shift)
+  # no transform field -> inherits transform_defaults (log_shift, labelled "log")
 - target_id: wk inc covid hosp
   metrics: [wis]
   transform:                 # replaces transform_defaults for this target
@@ -452,7 +453,7 @@ A target's single resolved transform applies uniformly to **every transformable 
 | `sample` | Not yet — support is under consideration |
 | `pmf` | No (categorical output types are not meaningfully transformed) |
 
-Metrics computed on a non-transformable output type (for example, the pmf metrics `log_score` or `rps`) are always scored on the natural scale and receive no transformed-scale column. See the [hubEvals metric × output-type compatibility reference](https://hubverse-org.github.io/hubEvals/reference/score_model_out.html#details) for which metrics apply to which output types.
+Metrics computed on a non-transformable output type (for example, the pmf metrics `log_score` or `rps`) are always scored on the natural scale and receive no transformed-scale column. See the [hubEvals `score_model_out()` documentation details](https://hubverse-org.github.io/hubEvals/reference/score_model_out.html#details) for the definition of each metric (e.g. what `wis` or `ae_median` mean) and which metrics (e.g. `wis`) can be applied to each output type (e.g. `quantile`).
 
 #### Opting a target out
 
@@ -467,6 +468,7 @@ transform_defaults:
   fun: log_shift
   args:
     offset: 1
+  label: log
 targets:
 - target_id: wk inc flu hosp
   metrics:
@@ -479,12 +481,12 @@ targets:
   transform: false
 ```
 
-In this example, `wk inc flu hosp` inherits `log_shift` (with `offset: 1`) from `transform_defaults`. The categorical target `wk flu hosp rate category` opts out explicitly with `transform: false`.
+In this example, `wk inc flu hosp` inherits `log_shift` (with `offset: 1`, labelled `log`) from `transform_defaults`, so its transformed-scale scores appear in the dashboard as e.g. `WIS (log)`. The categorical target `wk flu hosp rate category` opts out explicitly with `transform: false`.
 
 (predevals-transform-output)=
 #### How transformed scores appear in the dashboard
 
-When a transform applies to a target, the dashboard presents the transformed-scale metrics labelled with the transform's `label` in parentheses — for example, `WIS (log)` alongside `WIS`, or `Rel. WIS (log)` alongside `Rel. WIS`.
+When a transform applies to a target, the dashboard presents the transformed-scale metrics labelled with the transform's `label` in parentheses. In the example above, `label: log` means the transformed `WIS` appears as `WIS (log)` alongside the natural-scale `WIS` (and `Rel. WIS (log)` alongside `Rel. WIS`).
 
 <!-- TODO: add a screenshot of the evaluations table showing transformed-scale metric columns (e.g. `Rel. WIS (log)`) once the predevals dashboard UI for transforms is finalised, matching the screenshots earlier on this page. See review on hubverse-org/hubDocs#477. -->
 
--- a/docs/source/user-guide/dashboards.md
+++ b/docs/source/user-guide/dashboards.md
@@ -384,6 +384,84 @@ In principle, it would be possible to implement `round_filters` by specifying th
 
 :::
 
+(predevals-scale-transformations)=
+### Scale transformations (optional)
+
+For targets whose values span orders of magnitude (for example, count-based forecasts), it can be useful to evaluate predictions on a transformed scale rather than (or in addition to) the natural scale. PredEvals supports configuring scale transformations as part of `predevals-config.yml` from schema version `v1.1.0` onwards.
+
+To opt in, update your config's `schema_version` to the v1.1.0 URL:
+
+```yaml
+schema_version: https://raw.githubusercontent.com/hubverse-org/hubPredEvalsData/main/inst/schema/v1.1.0/config_schema.json
+```
+
+Transforms can be configured at two levels:
+
+- **`transform_defaults`** (top-level, optional): a default transformation applied to every target whose available output types support transformation.
+- **`targets[*].transform`** (per-target, optional): a transformation specific to one target. A per-target `transform` *replaces* `transform_defaults` entirely for that target. There is no merge between the two blocks.
+
+Each `transform` block, whether at the top level or per-target, has the following structure:
+
+| Property | Required | Description |
+| -------- | -------- | ----------- |
+| `fun`    | yes      | Name of the transform function. Must be one of the supported values listed below. |
+| `args`   | no       | Optional named arguments passed to `fun` (for example, `offset` for `log_shift`). Argument names must match the chosen function's formals. |
+| `append` | no       | If `true` (default), transformed-scale scores are produced in addition to natural-scale scores. If `false`, only transformed-scale scores are produced. |
+| `label`  | no       | Optional human-readable label associated with this transformation (for example, `log`). |
+
+#### Supported transform functions
+
+The schema restricts `fun` to the following functions, supported by the hubverse dashboards:
+
+| `fun`        | Source         | Description                                                              |
+| ------------ | -------------- | ------------------------------------------------------------------------ |
+| `log_shift`  | `scoringutils` | Natural log after adding a positive offset (set via `args.offset`). Useful when forecasts may be zero. |
+| `log1p`      | base R         | Natural log of `(1 + x)`. |
+| `log`        | base R         | Natural log (base e). |
+| `log10`      | base R         | Log base 10. |
+| `log2`       | base R         | Log base 2. |
+| `sqrt`       | base R         | Square root. |
+
+#### Opting a target out
+
+To exclude a single target from an inherited `transform_defaults` (without removing the default for other targets), set `transform: false` on that target.
+
+#### Example
+
+```yaml
+schema_version: https://raw.githubusercontent.com/hubverse-org/hubPredEvalsData/main/inst/schema/v1.1.0/config_schema.json
+rounds_idx: 0
+transform_defaults:
+  fun: log_shift
+  args:
+    offset: 1
+targets:
+- target_id: wk inc flu hosp
+  metrics:
+  - wis
+  - ae_median
+- target_id: wk flu hosp rate category
+  metrics:
+  - log_score
+  - rps
+  transform: false
+```
+
+In this example, `wk inc flu hosp` inherits `log_shift` (with `offset: 1`) from `transform_defaults`. The categorical target `wk flu hosp rate category` opts out explicitly with `transform: false`.
+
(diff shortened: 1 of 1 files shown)


docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,106 @@ In principle, it would be possible to implement `round_filters` by specifying th
...
+Scale transformations require schema version `v1.1.0` or later. If your config's `schema_version` points to an earlier version, update it to a version that supports transforms — for example, the one that introduced them:
+
+```yaml
+schema_version: https://raw.githubusercontent.com/hubverse-org/hubPredEvalsData/main/inst/schema/v1.1.0/config_schema.json
+```
+
+Transforms can be configured at two levels:
+
+- **`transform_defaults`** (top-level, optional): a default transformation applied to every target whose available output types support transformation.
+- **`targets[*].transform`** (per-target, optional): a transformation specific to one target. A per-target `transform` *replaces* `transform_defaults` entirely for that target.
```
lshandross: I don't really understand the syntax with the asterisk inside the brackets for the per-target transform field, and it wasn't clear to me what this looked like until an example config was shown further done on the page. Can we put an abbreviated toy config right below this block for clarity?

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,106 @@ In principle, it would be possible to implement `round_filters` by specifying th
...
+(predevals-transform-output)=
+#### How transformed scores appear in the dashboard
+
+When a transform applies to a target, the dashboard presents the transformed-scale metrics labelled with the transform's `label` in parentheses — for example, `WIS (log)` alongside `WIS`, or `Rel. WIS (log)` alongside `Rel. WIS`.
+
+- With **`append: true`** (the default), both the natural-scale and transformed-scale version of each affected metric are available.
+- With **`append: false`**, only the transformed-scale version is shown.
+
+Two metric families — interval coverage (for example `50% Cov.`, `95% Cov.`) and bias — are mathematically unchanged by any monotonic transform, so they are shown only once, with no separate transformed-scale variant.
+
```
lshandross, line 498: I think it would be beneficial to include a screenshot of what this actually looks like on the dashboard evaluations page, like the screenshots provided earlier on the page.

lshandross (changes requested):
Two small changes requested from me, but other than that this looks really good and matches the level of detail in the rest of this page!

docs/source/user-guide/dashboards.md
annakrystalli: Good call — the `targets[*].transform` notation was JSON-path jargon. I've replaced it with plain wording ("a `transform` field on an individual target") and added an abbreviated toy config right below the two-level explanation, showing all three behaviours: a target inheriting `transform_defaults`, one overriding it with its own transform, and one opting out via `transform: false`. Pushed in 38bf272.

docs/source/user-guide/dashboards.md
annakrystalli: Agreed that a screenshot would help. The catch is the predevals dashboard UI for transforms isn't finalised yet (this PR is intentionally landing ahead of the v1.1.0 release), so there's no real view to capture — I'd rather not mock one up and risk it not matching the shipped UI. I've added a `TODO` placeholder comment at that spot so we add a real screenshot (matching the other eval-page screenshots on this page) once the UI lands. Pushed in 38bf272 — does that work as a follow-up?

annakrystalli:
Thanks for the review @lshandross! Pushed 38bf272 addressing both points, plus a schema bump:

- **Per-target syntax (l.401):** dropped the `targets[*].transform` jargon for plain wording and added a small toy config showing inherit / override / opt-out.
- **Screenshot (l.473):** added a `TODO` placeholder rather than a mock-up, since the dashboard UI for transforms isn't finalised yet — replied in-thread, happy to do it as a follow-up once the UI lands.
- **Schema:** bumped all config-schema URLs on the page (main example, raw-schema link, docson widget, custom-evals example) from `v1.0.1` to `v1.1.0`, so the interactive schema widget now shows the transform properties.

Re-requesting your review when you have a moment.

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,129 @@ In principle, it would be possible to implement `round_filters` by specifying th
...
+```yaml
+schema_version: https://raw.githubusercontent.com/hubverse-org/hubPredEvalsData/main/inst/schema/v1.1.0/config_schema.json
+```
+
+Transforms can be configured at two levels:
+
+- **`transform_defaults`** (top-level, optional): a default transformation applied to every target whose available output types support transformation.
+- A **`transform`** field on an individual target (per-target, optional): a transformation specific to that one target, which *replaces* `transform_defaults` entirely for it.
+
+For example, the abbreviated config below sets a default transform, has the first target inherit it, the second target use its own transform instead, and the third opt out:
```
micokoch: As someone who doesn't use this, I would edit this slightly to clarify that you will provide details on what goes in this example, below.

```suggestion
Details regarding the structure and functions available for transforms will be provided below, but it may be easier to start with an illustrative case. In the example below, the abbreviated config sets a default transform, has the first target inherit it, the second target use its own transform instead, and the third opt out:
```

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,129 @@ In principle, it would be possible to implement `round_filters` by specifying th
 
 :::
 
+(predevals-scale-transformations)=
+### Scale transformations (optional)
```
micokoch: I would suggest adding a word that clarifies this is for evaluations. When I started reading this, I thought it referred to transformation of scores, like counts. Although it may be obvious to people who read this whole page from start to finish, it is less so, for people who are quickly skimming the manual (as most of us do). Maybe the title below, or something like it?

```suggestion
### Scale transformations for PredEvals (optional)
```

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,129 @@ In principle, it would be possible to implement `round_filters` by specifying th
...
+transform_defaults:          # applies to all transformable targets by default
+  fun: log_shift
+  args:
+    offset: 1
+targets:
+- target_id: wk inc flu hosp
+  metrics: [wis]
+  # no transform field -> inherits transform_defaults (log_shift)
+- target_id: wk inc covid hosp
+  metrics: [wis]
```
micokoch, line 416: Just a thought, but as I was thinking about this example, I wondered why you would apply different transformations for the same metric, but with different `target_id`s (it sounded fishy). Maybe this is from a use case, and so my comment is wrong, but for this second case, could you perhaps do `wk inc flu death` and `[ae_median]`? Or maybe that example is also wrong?

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,129 @@ In principle, it would be possible to implement `round_filters` by specifying th
 
 :::
 
+(predevals-scale-transformations)=
+### Scale transformations (optional)
+
```
micokoch, line 389: This is not specific to the new content, but in general, in this section, and in fact in none of the hubDocs, was I able to find an explanation of some of the metrics. Specifically, I was searching for what the "ae" stood for in `ae_median`. I searched in hubDocs, and not finding anything, went to Google. Reading "absolute error" made it seem obvious, but it would be nice for it to be explicit in this documentation. Or at the very least, to have a pointer to where we can read those definitions (I didn't find one). In yesterday's presentation I saw how the dashboards now have "metric definitions", could we include something like that, here? Or at least a link to learn more?

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,129 @@ In principle, it would be possible to implement `round_filters` by specifying th
...
+- **`transform_defaults`** (top-level, optional): a default transformation applied to every target whose available output types support transformation.
+- A **`transform`** field on an individual target (per-target, optional): a transformation specific to that one target, which *replaces* `transform_defaults` entirely for it.
+
+For example, the abbreviated config below sets a default transform, has the first target inherit it, the second target use its own transform instead, and the third opt out:
+
+```yaml
+transform_defaults:          # applies to all transformable targets by default
+  fun: log_shift
+  args:
+    offset: 1
```
micokoch, line 409: Sorry if I'm being pedantic, but could you choose a different offset than "1" (maybe 0.5, or 10)? The reason is that you can use `log1p` instead, so why would you do it the long way of writing down the offset.

docs/source/user-guide/dashboards.md
```diff
@@ -384,6 +384,129 @@ In principle, it would be possible to implement `round_filters` by specifying th
...
+
+A target's single resolved transform applies uniformly to **every transformable output type** the target declares. The transformable output types are:
+
+| Output type | Transformable? |
+| ----------- | -------------- |
+| `mean`, `median`, `quantile` | Yes |
+| `sample` | Not yet — support is under consideration |
+| `pmf` | No (categorical output types are not meaningfully transformed) |
+
+Metrics computed on a non-transformable output type (for example, the pmf metrics `log_score` or `rps`) are always scored on the natural scale and receive no transformed-scale column. See the [hubEvals metric × output-type compatibility reference](https://hubverse-org.github.io/hubEvals/reference/score_model_out.html#details) for which metrics apply to which output types.
```
micokoch:
```suggestion
Metrics computed on a non-transformable output type (for example, the pmf metrics `log_score` or `rps`) are always scored on the natural scale and receive no transformed-scale column. See the [hubEvals `score_model_out` documentation details](https://hubverse-org.github.io/hubEvals/reference/score_model_out.html#details) to see what metrics (e.g., `wis`) can be applied to each `output-type` (e.g., `quantile`).
```

micokoch (approved):
This is great Anna! It was very clear and well explained, even for me, who has never used any of these tools, including the dashboard (other than as a consumer). I left suggestions, but you are of course welcome to ignore them. I didn't mean to be nitpicky with my comments, but for someone coming at this from a relatively naïve perspective, I thought the points below could help the ideas easier to follow. Thanks for doing it!

docs/source/user-guide/dashboards.md
lshandross: Yes, that's totally fine, thank you!

lshandross (approved):

docs/source/user-guide/dashboards.md
annakrystalli: Good suggestion — I've reworked the lead-in so it signals the structure and available functions are detailed below, and frames the config as a starting illustration. Thanks!

docs/source/user-guide/dashboards.md
annakrystalli: Agreed — scoping it to PredEvals in the heading helps skimmers land in the right place. Updated the heading to **"Scale transformations for PredEvals (optional)"**.

docs/source/user-guide/dashboards.md
annakrystalli: Thanks for flagging this — it's a fair instinct. The repeated `wis` is actually deliberate: I wanted the *only* thing varying between the first two targets to be the `transform` field, so the example reads purely as "this is how a per-target transform overrides the default," independent of which metric is in play. If I changed both the target and the metric, it could imply the transform choice is somehow tied to the metric, which isn't the case. So I've kept it as-is, but let me know if you feel it still reads as misleading.

docs/source/user-guide/dashboards.md
annakrystalli: Good point that the abbreviations deserve a definition. The page already introduces these a bit earlier — the worked example at the top of the PredEvals section spells out "the weighted interval score (WIS), absolute error of the median…" — and I've expanded the link below the output-types table to point explicitly at the [hubEvals `score_model_out()` reference](https://hubverse-org.github.io/hubEvals/reference/score_model_out.html#details) (e.g. what `wis` / `ae_median` mean) and the [scoringutils scoring-rules article](https://epiforecasts.io/scoringutils/articles/scoring-rules.html). We're deliberately keeping the metric definitions in hubEvals/scoringutils rather than copying them here — these are the dashboard-specific docs, and the metric details track scoringutils, so a second copy would just be another place to keep in sync.

docs/source/user-guide/dashboards.md
annakrystalli: I'd actually like to keep `offset: 1` here. The purpose of this example is to show how `args` are passed to a `fun`, and `log(x + 1)` is one of the most common, realistic choices for count data that can hit zero. The fact that `log1p` happens to be an equivalent shortcut doesn't really change what the example is teaching — `log_shift` is the general, flexible function and `args` is the thing being demonstrated. So I've left it as `offset: 1`. Appreciate the close read though!

docs/source/user-guide/dashboards.md
annakrystalli: Adopted, thank you — that's much more concrete. I've reworded it to point at the `score_model_out()` details for the definition of each metric (e.g. what `wis` or `ae_median` mean) and which metrics can be applied to each output type. This also doubles as the pointer for your metric-definitions comment above.

annakrystalli:
Thanks so much for the thorough read, @micokoch — a reviewer coming at this fresh is exactly the perspective these docs need. I've pushed changes addressing your suggestions: clearer section heading and example lead-in, a more concrete link to the hubEvals metric reference, and an explicit `label` in the examples so the `WIS (log)` naming is traceable. I've left the per-target example metric and the `offset: 1` as-is (replied inline on those two with the reasoning), but happy to revisit if you feel strongly. Much appreciated!

micokoch:
Sounds good to me. Great work @annakrystalli !

PR #485, opened 2026-06-17
Add CLAUDE.md file and update first admonition

micokoch:
Add CLAUDE.md file and update banner on main page, so that it points to hubverse site but removing talk of new.


matthewcornell (approved):
LGTM!

PR #494, opened 2026-07-09
Complete author list in hubEnsembles BibTeX citation

micokoch:
The BibTeX entry for the hubEnsembles paper in `docs/source/overview/cite.md` previously truncated the author list with `and others` after the sixth author (Consortium of Infectious Disease Modeling Hubs), omitting Nicholas G. Reich and Evan L. Ray.

This replaces `and others` with the two remaining authors so the BibTeX entry lists the complete author list, matching the published paper (doi:10.1002/sim.70333) and the CrossRef record:

Shandross, Howerton, Contamin, Hochheiser, Krystalli, Consortium of Infectious Disease Modeling Hubs, Reich, Ray.

The APA and Vancouver renderings on the same page intentionally keep their `et al.` truncation per their respective style guides; only the BibTeX entry, which should carry the full list for reuse, is updated.


LucieContamin (approved):

matthewcornell (approved):
LGTM

PR #498, opened 2026-07-15
Document that the hubverse accepts dates, not datetimes

### LINKED ISSUE #177: Add section about date columns
A recent bug report in [#118](https://github.com/hubverse-org/hubValidations/issues/118) where folks submitting datetime columns where they should have been submitting dates was causing an execution error (fixed in https://github.com/hubverse-org/hubValidations/pull/119) prompted me to notice that:
- Apart from the schema for specific task IDs we don't really mention that we currently only support dates, specifically in ISO format and not datetimes across the hubverse.
- There is also nothing stopping admins from adding their own custom taskIDs where they might expect data times (although it's hard to imagine how they would encode datetime valid values in the config). In any case, our framework currently cannot support handling of datetimes. This would require us to add mappings between json datetime data type to R datetime datatype and finally to the arrow datetime datatype equivalent.

So my questions are:

1. Should we add a specific section on dates somewhere in `hubDocs`? Any preferences for where?
2. Are datetimes sth we should support? If not, should we make that clear in the aforementioned docs section?

_Originally posted by @annakrystalli in https://github.com/hubverse-org/.github/discussions/31_

The initial thought in the meeting was to documentation that we specify only accepting dates for now in the documentation.

Possibly in:

 - this repository
 - hubValidations
 - hubData (maybe)

Willing to hear about other options.

micokoch:
## Document that the hubverse accepts dates, not datetimes

Closes #177

### Background

A bug report ([hubValidations#118](https://github.com/hubverse-org/hubValidations/issues/118)) where modelers submitted datetime columns in place of dates surfaced that our docs don't clearly state that the hubverse currently supports **dates only** (ISO 8601, `YYYY-MM-DD`), not datetimes. Our framework cannot currently handle datetimes, so this PR documents the limitation rather than adding support for it.

Placement follows the discussion in #177 (and the linked [team discussion](https://github.com/hubverse-org/.github/discussions/31)): keep the primary note focused on task ID data types, and reiterate the limitation for model output files as a whole. No changes were made to hubValidations or hubData, per that discussion.

### Changes

- **`docs/source/user-guide/tasks.md`** — new `#### Date values in task ID variables` subsection (anchor `task-id-data-types`) at the end of the Task ID variables section. States that date-valued task IDs must be ISO 8601 (`YYYY-MM-DD`) strings, that datetimes are not supported anywhere in the framework, and that sub-daily resolution should be represented with a separate task ID variable.
- **`docs/source/user-guide/model-output.md`** — note in "Details about model output column specifications" reiterating the dates-only limitation across all model output columns, cross-linked to the tasks.md note.

### Notes for reviewers

- Used in-section `important`/`note` admonitions (consistent with these pages) rather than the footnote style originally floated in the issue.
- `sphinx build` succeeds locally with no new warnings; the new anchor and cross-reference resolve.

### DIFF
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -182,6 +182,17 @@ We therefore strongly suggest that Hubs adopt the following standard task ID or 
 [^new-vars]: As Hubs define new modeling tasks, they may need to introduce new task ID variables that have not been used before.
 In those cases, the new variables should be added to this list to ensure that the concepts are documented centrally and can be reused in future efforts.
 
+(task-id-data-types)=
+#### Date values in task ID variables
+
+Several standard task ID variables represent dates (for example, `origin_date`{.codeitem}, `forecast_date`{.codeitem}, `reference_date`{.codeitem}, and `target_date`{.codeitem}/`target_end_date`{.codeitem}).
+
+:::{important}
+Date-valued task IDs must be represented as strings in ISO 8601 date format (`YYYY-MM-DD`), for example `"2024-11-07"`.
+
+The hubverse currently supports **dates only, not datetimes**. There is no way to attach a time component (e.g., `2024-11-07T14:30:00`) to a task ID value, and datetimes are not supported anywhere in the hubverse framework. If a modeling task requires finer-than-daily resolution, represent the time component using a separate task ID variable.
+:::
+
 (output-types)=
 ## Output types
 
--- a/docs/source/user-guide/tasks.md
+++ b/docs/source/user-guide/tasks.md
@@ -193,17 +193,16 @@ Date-valued task IDs must be represented as strings in ISO 8601 date format (`YY
 The hubverse currently supports **dates only, not datetimes**. There is no datetime type and no time-of-day type anywhere in the hubverse framework, so you cannot attach a time component (e.g., `2024-11-07T14:30:00`) to a task ID value.
 :::
 
-If a modeling task requires finer-than-daily resolution, split the sub-daily component into its own task ID variable. The benefit is *not* that this represents time "properly"; it is that the component can then be encoded as a plain `integer` or `character` value (for example, `hour` as `14`, or `"14:30"`) — data types the hubverse already reads and validates (via its Arrow/R tooling). No time data type is needed, and this works today with no changes to hubData or hubValidations.
+If a modeling task appears to require finer-than-daily resolution, we currently recommend avoiding it where possible. The hubverse has no time or datetime type, so there is no first-class way to represent sub-daily resolution, and the available workaround carries real validation costs (see below). Consider whether a daily-or-coarser formulation can meet your needs first.
+
+If sub-daily resolution is genuinely unavoidable, the only option today is to encode the sub-daily component as a separate task ID variable with a plain `integer` or `character` type (for example, `hour` as `14`, or `"14:30"`). These are types the hubverse already reads and validates via its Arrow/R tooling, with no changes to hubData or hubValidations. Weigh the following before doing so.
 
 :::{warning}
-The hubverse read and validation functions treat such a variable **strictly as an `integer` or `character` column, with no temporal semantics**: no timezone, no time-of-day type, no interpretation of its ordering as time, and no validity checking (for example, `hour` values of `47`, or `"25:99"`, would pass validation unless the hub adds its own custom validation checks). Any temporal meaning lives entirely in the hub's config and the modeler's convention.
+**No temporal semantics.** Such a variable is treated strictly as an `integer` or `character` column: no timezone, no time-of-day type, and no interpretation of its ordering as time. Values are validated like any other task ID, by membership against the accepted values you enumerate in the hub's config, not against any notion of a valid time. An out-of-range value such as `hour = 47` is only rejected if it falls outside the accepted values you have defined.
 
-To keep conventions compatible across hubs and avoid sorting surprises, adopt a canonical encoding — either:
+**Validation performance cost.** The hubverse validates value combinations against an expanded grid, roughly the cartesian product of every task ID's accepted values within a modeling task. A temporal task ID multiplies the size of that grid by its number of accepted values: an hourly variable (24 values) multiplies it about 24 times, while a minute-level variable (1440 values) multiplies it about 1440 times, which can make validation impractically slow and memory-heavy. Keep the resolution as coarse as possible.
 
-- an **`integer`** in a fixed range (e.g., `hour` from `0` to `23`), or
-- a **zero-padded, fixed-width `character` string** (e.g., `"HH:MM"`, so `"09:30"` rather than `"9:30"`).
-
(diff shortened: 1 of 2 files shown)


docs/source/user-guide/tasks.md
```diff
@@ -182,6 +182,17 @@ We therefore strongly suggest that Hubs adopt the following standard task ID or
...
+(task-id-data-types)=
+#### Date values in task ID variables
+
+Several standard task ID variables represent dates (for example, `origin_date`{.codeitem}, `forecast_date`{.codeitem}, `reference_date`{.codeitem}, and `target_date`{.codeitem}/`target_end_date`{.codeitem}).
+
+:::{important}
+Date-valued task IDs must be represented as strings in ISO 8601 date format (`YYYY-MM-DD`), for example `"2024-11-07"`.
+
+The hubverse currently supports **dates only, not datetimes**. There is no way to attach a time component (e.g., `2024-11-07T14:30:00`) to a task ID value, and datetimes are not supported anywhere in the hubverse framework. If a modeling task requires finer-than-daily resolution, represent the time component using a separate task ID variable.
+:::
```
annakrystalli, line 206: Given we don't support a time data type either, I think we need to be explicit about *why* this helps, because on its own the phrasing might mislead admins/modelers.

The benefit isn't that a separate task ID lets us represent time properly but that splitting the sub-daily component into its own task ID means it *can* be encoded as a plain `integer` or `character` (e.g. `hour = 14`, or `"14:30"`), types our Arrow↔R read path and validation already handle. So no time type is needed at all, and it's implementable today with no changes to hubData/hubValidations.

But I feel we should be explicit about this or it invites the wrong expectation. Specifically: the hubverse read and validation functions treat such variables strictly as `integer` or `character`, with **no** temporal semantics. No time-of-day type, no timezone, no ordering-as-time, no validity checking (`hour = 47` or `"25:99"` would pass unless custom validation functions where implemented). Any temporal meaning lives in the hub's config and the modeler's convention.

Given that, could we add a recommended canonical encoding (e.g. integer hour `0` to `23`, or zero-padded `"HH:MM"`)? Without one, hubs will pick incompatible conventions, and since these are untyped columns, lexical vs numeric sorting will bite (`"9"` sorts after `"10"`).

annakrystalli (commented):
Nice work @micokoch and thanks for tackling this! I have one question/suggestion before approving, see inline comment.

docs/source/user-guide/tasks.md
micokoch: Thanks, @annakrystalli - this was good feedback. I updated the PR:

- Reworked the note to explain *why* a separate task ID works: the sub-daily component is encoded as a plain `integer`/`character`, types the Arrow/R read path and validation already handle — no time type, no hubData/hubValidations changes.
- Added an explicit warning that these columns have no temporal semantics (no timezone, no time-of-day type, no ordering-as-time, no validity checking; e.g. `hour = 47` or `"25:99"` pass unless the hub adds custom checks).
- Added a recommended canonical encoding (integer `0`–`23`, or zero-padded `"HH:MM"`) and noted the lexical-sort pitfall (`"9"` after `"10"`).

Let me know if you need additional changes.

docs/source/user-guide/tasks.md
```diff
@@ -182,6 +182,30 @@ We therefore strongly suggest that Hubs adopt the following standard task ID or
...
+:::{warning}
+The hubverse read and validation functions treat such a variable **strictly as an `integer` or `character` column, with no temporal semantics**: no timezone, no time-of-day type, no interpretation of its ordering as time, and no validity checking (for example, `hour` values of `47`, or `"25:99"`, would pass validation unless the hub adds its own custom validation checks). Any temporal meaning lives entirely in the hub's config and the modeler's convention.
+
+To keep conventions compatible across hubs and avoid sorting surprises, adopt a canonical encoding — either:
+
+- an **`integer`** in a fixed range (e.g., `hour` from `0` to `23`), or
+- a **zero-padded, fixed-width `character` string** (e.g., `"HH:MM"`, so `"09:30"` rather than `"9:30"`).
+
+Because these columns carry no temporal type, character values are sorted lexically (character by character) rather than numerically — so without zero-padding, `"9"` sorts *after* `"10"`. Prefer zero-padded strings or integers to avoid this.
+:::
```
annakrystalli: Following up on our thread: rather than showing how to add sub-daily resolution, I think we should **discourage** it for now and be explicit about the cost. This reframe leads with "avoid where possible", corrects the "no validity checking" point (values do validate, as set membership against the accepted values enumerated in config), and surfaces the real blocker: adding a temporal task ID multiplies the expanded validation grid by its cardinality (minute-level ~1440x), which quickly makes validation impractical. Suggested rewording:

```suggestion
If a modeling task appears to require finer-than-daily resolution, we currently recommend avoiding it where possible. The hubverse has no time or datetime type, so there is no first-class way to represent sub-daily resolution, and the available workaround carries real validation costs (see below). Consider whether a daily-or-coarser formulation can meet your needs first.

If sub-daily resolution is genuinely unavoidable, the only option today is to encode the sub-daily component as a separate task ID variable with a plain `integer` or `character` type (for example, `hour` as `14`, or `"14:30"`). These are types the hubverse already reads and validates via its Arrow/R tooling, with no changes to hubData or hubValidations. Weigh the following before doing so.

:::{warning}
**No temporal semantics.** Such a variable is treated strictly as an `integer` or `character` column: no timezone, no time-of-day type, and no interpretation of its ordering as time. Values are validated like any other task ID, by membership against the accepted values you enumerate in the hub's config, not against any notion of a valid time. An out-of-range value such as `hour = 47` is only rejected if it falls outside the accepted values you have defined.

**Validation performance cost.** The hubverse validates value combinations against an expanded grid, roughly the cartesian product of every task ID's accepted values within a modeling task. A temporal task ID multiplies the size of that grid by its number of accepted values: an hourly variable (24 values) multiplies it about 24 times, while a minute-level variable (1440 values) multiplies it about 1440 times, which can make validation impractically slow and memory-heavy. Keep the resolution as coarse as possible.

**Encoding.** These columns carry no temporal type, so character values sort lexically rather than numerically (without zero-padding, `"9"` sorts after `"10"`). If you do introduce one, use a canonical encoding: an `integer` in a fixed range (for example, `0` to `23`), or a zero-padded, fixed-width `character` string (for example, `"09:30"` rather than `"9:30"`).
:::
```

annakrystalli (changes requested):
Thanks for these edits @micokoch, they really tightened things up! While re-reading I had a bit of a lightbulb moment about how our validation actually works (values validate as set membership against config, and combinations expand into a grid that a sub-daily task ID would blow up), and it shifted how I think we should frame this. I have left a suggestion on the paragraph and warning that flips it towards discouraging sub-daily resolution. Happy to discuss!

docs/source/user-guide/tasks.md
micokoch: This looks great, thanks!

annakrystalli (approved):

micokoch:
Hi Anna, your fixes look great and make complete sense - thank you! I think it is all clearer now (for me and for the documentation).

PR #508, opened 2026-07-30
Add data-revision warning for retrospective forecasting and evaluation

micokoch:
Adds a prominent warning that hub data must be used carefully for retrospective forecast generation and evaluation, since surveillance data is revised over time. Comes out of a suggestion from the Delphi group; wording was discussed and agreed in this [discussion](https://github.com/orgs/hubverse-org/discussions/45).

Changes:
- `user-guide/target-data.md`: full warning admonition (anchor `data-revisions-warning`), plus a `seealso` back-link from the `as_of` versioning section.
- `developer/dashboard-predevals.md`: short warning on the evaluations dashboard page.
- `user-guide/dashboards.md`: inline cross-reference where the PredTimeChart section discusses showing the as-of data version.

All cross-references point to the single canonical anchor and resolve in a clean `sphinx-build`.

### DIFF
--- a/docs/source/user-guide/target-data.md
+++ b/docs/source/user-guide/target-data.md
@@ -28,7 +28,7 @@ document.
 :class: warning
 :name: data-revisions-warning
 
-Forecasts in a real-time hub were made using the data **available at the time**. Surveillance data is often revised after its initial release (reporting delays, backfill, and corrections), so the observed value for a given target date can change in later data versions. In practice, making predictions using finalized data and then evaluated with that same data can make retrospective performance look substantially better than it really was in real time.
+Forecasts in a real-time hub were made using the data **available at the time**. Surveillance data is often revised after its initial release (reporting delays, backfill, and corrections), so the observed value for a given target date can change in later data versions. In practice, generating forecasts from finalized data and then evaluating them against that same data can make retrospective performance look substantially better than it really was in real time.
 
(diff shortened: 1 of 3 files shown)


docs/source/user-guide/target-data.md
```diff
@@ -24,6 +24,23 @@ document.
 [^truth]: Time series data is sometimes referred to as "ground truth" data, but
     we no longer use this term in the hubverse.
 
+:::{admonition} Data revisions, retrospective forecasting, and evaluation
+:class: warning
+:name: data-revisions-warning
+
+Forecasts in a real-time hub were made using the data **available at the time**. Surveillance data is often revised after its initial release (reporting delays, backfill, and corrections), so the observed value for a given target date can change in later data versions. In practice, evaluating on finalized data can make retrospective performance look substantially better than it really was in real time.
```
lshandross:
```suggestion
Forecasts in a real-time hub were made using the data **available at the time**. Surveillance data is often revised after its initial release (reporting delays, backfill, and corrections), so the observed value for a given target date can change in later data versions. In practice, making predictions using finalized data and then evaluated with that same data can make retrospective performance look substantially better than it really was in real time.
```

The original way the last sentence in this paragraph was worded wasn't quite accurate, so I tried to revise based on what I think the intended meaning was.

lshandross (approved):
One suggestion/comment about re-wording but looks good other than that. This is an important warning to include!

